PR #2, opened 2021-03-23
Create debian packages for grub-trenchboot_support

Cezarus27:
These changes provide files that are necessary to build Debian packages for GRUB with trenchboot support.
It is a modification of files from Debian GRUB maintainers from https://salsa.debian.org/grub-team/grub .

For now, it works on a separate branch called "debian-upstream". This branch is synchronizing when new commits appear at the main `trenchboot-support-2.04` branch. If this condition is true, then `debian-upstream` branch fetch changes from `trenchboot-support-2.04`, and then the CI pipeline starts up.


pietrushnic:
@macpijan @Cezarus27 is it really good way of doing things 200k SLOC? Can't we rely on external source code to be fetched for building packages?

PR #6, opened 2022-09-01
Intel txt mb2

miczyg1:


grub-core/loader/multiboot.c
```diff
@@ -164,7 +164,22 @@ efi_boot (struct grub_relocator *rel __attribute__ ((unused)),
 static void
 normal_boot (struct grub_relocator *rel, struct grub_relocator32_state state)
 {
-  state.edi = SLP_NONE;
+  struct grub_slaunch_params *slparams = grub_slaunch_params();
```
krystian-hebel, line 169: This probably will break builds without slaunch module.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -120,11 +171,17 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
           return err;
         }
 
-      mld->load_base_addr = get_physical_target_address (ch);
-      source = get_virtual_current_address (ch);
+      mld->load_base_addr = get_physical_target_address (ch) + slparams->mle_ptab_size;
```
krystian-hebel: This breaks alignment if original `mld->align` is not a multiply of `slparams->mle_ptab_size`.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -106,12 +115,54 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+	    total_size = ALIGN_UP(total_size, GRUB_TXT_PMR_ALIGN);
+
+	    slparams->mle_size = total_size;
+
+	    slparams->mle_ptab_size = grub_txt_get_mle_ptab_size (total_size);
+	    slparams->mle_ptab_size = ALIGN_UP (slparams->mle_ptab_size, GRUB_TXT_PMR_ALIGN);
+
+	    total_size += slparams->mle_ptab_size;
+	    /* Do not go below GRUB_TXT_PMR_ALIGN. */
+	    if (mld.align > GRUB_TXT_PMR_ALIGN)
```
krystian-hebel:
```suggestion
	    if (mld->align > GRUB_TXT_PMR_ALIGN)
```
Here and every other use of `mld`

grub-core/loader/multiboot_mbi2.c
```diff
@@ -390,6 +500,46 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
...
+	slparams->tpm_evt_log_size = GRUB_SLAUNCH_TPM_EVT_LOG_SIZE;
+
+	grub_memset (get_virtual_current_address (ch), 0, slparams->tpm_evt_log_size);
+
+	grub_dprintf ("multiboot_loader", "tpm_evt_log_base = %lx, tpm_evt_log_size = %x\n",
+		      (unsigned long) slparams->tpm_evt_log_base,
+		      (unsigned) slparams->tpm_evt_log_size);
+
+	if (grub_relocator_alloc_chunk_align (relocator, &ch, 0x1000000,
+						    0xffffffff & ~slparams->mle_ptab_size,
```
krystian-hebel: Is this a fancy optimization or protection against underflow?

grub-core/loader/multiboot_mbi2.c
```diff
@@ -708,21 +858,34 @@ grub_multiboot2_make_mbi (grub_uint32_t *target)
...
 
-  err = grub_relocator_alloc_chunk_align (grub_multiboot2_relocator, &ch,
-					  MBI_MIN_ADDR, 0xffffffff - bufsize,
-					  bufsize, MULTIBOOT_TAG_ALIGN,
-					  GRUB_RELOCATOR_PREFERENCE_NONE, 1);
+  if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+    {
+      /* MBI was saved and allocated earlier for slaunch */
+      *target = slparams->boot_params_addr;
+      /* For x86 and x64 the allocated physical addr of chunk is the same as virtual */
```
krystian-hebel: Proof? IIRC it's only the same if it can be allocated at target address, otherwise it can be different and moved by relocator just before jump to kernel.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -196,7 +202,20 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
 	      }
 	  break;
 	}
-	       
+      case MULTIBOOT_HEADER_TAG_END:
```
krystian-hebel: This will never be matched due to `tag->type != MULTIBOOT_TAG_TYPE_END;` in encapsulating `for`.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -196,7 +202,20 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
...
+	if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+	  {
+	    /* MLE header will be right after MB2 header */
+	    mle_header_off = ALIGN_UP((grub_addr_t)tag + sizeof(*tag), 16);
+	    if ( grub_memcmp (mle_header_off, GRUB_TXT_MLE_UUID, 16) )
+	      {
+		grub_free (mld.buffer);
+		return grub_error (GRUB_ERR_BAD_ARGUMENT, "MLE header not found");
+	      }
+	    slparams->mle_header_offset = mld.buffer - mle_header_off;
```
krystian-hebel:
```suggestion
	    slparams->mle_header_offset = mle_header_off - mld.buffer;
```

grub-core/loader/multiboot.c
miczyg1: Linux likewise... But that should not be a concern in this particular moment.

grub-core/loader/multiboot_elfxx.c
miczyg1: I have done alignment checks later. But maybe not for every path yet...

grub-core/loader/multiboot_mbi2.c
miczyg1: See GRUB code, it uses the same address when accessing the virtual and physical addresses of the chunk structure.

grub-core/loader/multiboot_mbi2.c
miczyg1: Idea copied from Linux implementation

grub-core/loader/multiboot_mbi2.c
krystian-hebel: https://github.com/TrenchBoot/grub/blob/master/grub-core/lib/relocator.c#L92
https://github.com/TrenchBoot/grub/blob/master/grub-core/lib/relocator.c#L98
https://github.com/TrenchBoot/grub/blob/master/grub-core/lib/relocator.c#L1467 and thereabouts.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -196,7 +202,6 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
 	      }
 	  break;
 	}
-
```
krystian-hebel, line 199: Unnecessary whitespace change.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -268,6 +273,27 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
 	break;
       }
 
+  if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+    {
+      tag = (struct multiboot_header_tag *) ((grub_uint32_t *) tag + ALIGN_UP (tag->size, MULTIBOOT_TAG_ALIGN) / 4);
```
krystian-hebel: No need for this line, it is already aligned by iteration expression of `for`.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -268,6 +273,27 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
...
+      if (tag->type == MULTIBOOT_HEADER_TAG_END)
+        {
+	  /* MLE header will be right after MB2 header */
+	  mle_header_off = ALIGN_UP((grub_addr_t)tag + sizeof(*tag), 16);
+	  if ( grub_memcmp ((void *)mle_header_off, GRUB_TXT_MLE_UUID, 16) )
+	    {
+		grub_free (mld.buffer);
+		return grub_error (GRUB_ERR_BAD_ARGUMENT, "MLE header not found");
+	    }
+	  slparams->mle_header_offset = (grub_addr_t)mld.buffer - (grub_addr_t)mle_header_off;
```
krystian-hebel: This is reversed. You're subtracting bigger value from a smaller one.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -295,29 +321,120 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
...
+	      slparams->mle_size = total_size;
+
+	      slparams->mle_ptab_size = grub_txt_get_mle_ptab_size (total_size);
+	      slparams->mle_ptab_size = ALIGN_UP (slparams->mle_ptab_size, GRUB_TXT_PMR_ALIGN);
+
+	      total_size += slparams->mle_ptab_size;
+	      /* Do not go below GRUB_TXT_PMR_ALIGN. */
+	      if (mld.align > GRUB_TXT_PMR_ALIGN)
+		{
+		  mld.min_addr = (mld.min_addr > slparams->mle_ptab_size) ?
```
krystian-hebel: Isn't it required by TXT for page tables to be located below MLE?

grub-core/loader/multiboot_elfxx.c
```diff
@@ -106,12 +115,54 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
 		    (long) mld->align, mld->preference, load_size,
 		    mld->avoid_efi_boot_services);
 
-      if (load_size > mld->max_addr || mld->min_addr > mld->max_addr - load_size)
+      if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
```
krystian-hebel, line 114: The same block is present in `multiboot2_mbi.c`, are both required?

grub-core/loader/multiboot_mbi2.c
```diff
@@ -295,29 +321,120 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
...
+
+	      slparams->mle_size = total_size;
+
+	      slparams->mle_ptab_size = grub_txt_get_mle_ptab_size (total_size);
+	      slparams->mle_ptab_size = ALIGN_UP (slparams->mle_ptab_size, GRUB_TXT_PMR_ALIGN);
+
+	      total_size += slparams->mle_ptab_size;
+	      /* Do not go below GRUB_TXT_PMR_ALIGN. */
+	      if (load_addr < slparams->mle_ptab_size || 
+	          ((load_addr > slparams->mle_ptab_size) &&
```
krystian-hebel: This misses the case when `load_addr == slparams->mle_ptab_size`.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -295,29 +321,120 @@ grub_multiboot2_load (grub_file_t file, const char *filename)
...
-						  code_size, mld.align ? mld.align : 1,
+						  mld.min_addr, mld.max_addr - total_size,
+						  total_size, mld.align ? mld.align : 1,
 						  mld.preference, keep_bs);
 	}
       else
+	{
+	  if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+	    {
+	      if (!IS_ALIGNED(load_addr, GRUB_TXT_PMR_ALIGN))
```
krystian-hebel: Just to confirm if I understand it correctly:
1. page tables are first in the allocated chunk, they **do not** have to be aligned to PMR
2. they are immediately followed (without any padding) by kernel that is aligned at least to PMR
3. kernel is followed by MBI, aligned to 8B or whatever the tags require
4. end of chunk is aligned up to PMR

Then MLE is set to everything between 2-4, including padding at the end, and that is measured to PCR17 by ACM, right?

grub-core/loader/i386/txt/txt.c
```diff
@@ -576,7 +576,7 @@ init_txt_heap (struct grub_slaunch_params *slparams, struct grub_txt_acm_header
   os_sinit_data->mle_ptab = slparams->mle_ptab_target;
   os_sinit_data->mle_size = slparams->mle_size;
 
-  os_sinit_data->mle_hdr_base = slparams->mle_header_offset;
+  os_sinit_data->mle_hdr_base = slparams->mle_start + slparams->mle_header_offset;
```
krystian-hebel: Is it used by Linux path?

grub-core/loader/i386/txt/txt.c
```diff
@@ -599,14 +631,30 @@ init_txt_heap (struct grub_slaunch_params *slparams, struct grub_txt_acm_header
...
+  if (grub_get_tpm_ver () == GRUB_TPM_20)
+    {
+      /* CBnT bits 5:4 must be 11b, since D/A mapping is the only one supported. */
+      if ((sinit_caps & os_sinit_data->capabilities) != os_sinit_data->capabilities)
+        return grub_error (GRUB_ERR_BAD_ARGUMENT,
+               N_("Details/authorities PCR usage is not supported"));
+    }
+  else
+    {
+      if ((sinit_caps & GRUB_TXT_CAPS_TPM_12_AUTH_PCR_USAGE) != GRUB_TXT_CAPS_TPM_12_AUTH_PCR_USAGE)
```
krystian-hebel: Doing mask and compare doesn't make sense for single bits. This can be simplified to:
```suggestion
      if (!(sinit_caps & GRUB_TXT_CAPS_TPM_12_AUTH_PCR_USAGE))
```

grub-core/loader/i386/txt/txt.c
```diff
@@ -599,14 +631,30 @@ init_txt_heap (struct grub_slaunch_params *slparams, struct grub_txt_acm_header
...
+	{
+	  grub_dprintf ("slaunch", "Details/authorities PCR usage is not supported. Trying legacy");
+	  if ((sinit_caps & GRUB_TXT_CAPS_TPM_12_NO_LEGACY_PCR_USAGE) == GRUB_TXT_CAPS_TPM_12_NO_LEGACY_PCR_USAGE)
```
krystian-hebel: Same as above.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -101,30 +111,99 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+	  /*
+	   * We allocate the the binary together with page tables and MBI
+	   * to make one contiguous block for MLE. We have to align up to
+	   * PMR (2MB).
+	   */
+	  total_size = ALIGN_UP(load_size, 8); // MULTIBOOT_TAG_ALIGN
+	  total_size += GRUB_MULTIBOOT (get_mbi_size)();
+	  total_size = ALIGN_UP(total_size, GRUB_TXT_PMR_ALIGN);
+
+	  slparams->mle_size = total_size;
```
krystian-hebel, line 125: We probably shouldn't measure MBI together with code.
```suggestion
	  slparams->mle_size = load_size;
```

grub-core/loader/multiboot_elfxx.c
```diff
@@ -101,30 +111,99 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+				(mld->min_addr - slparams->mle_ptab_size) : mld->align;
+	      mld->min_addr = ALIGN_UP (mld->min_addr, mld->align);
+	    }
+	  else
+	    {
+	      mld->min_addr = (mld->min_addr > slparams->mle_ptab_size) ?
+				(mld->min_addr - slparams->mle_ptab_size) : GRUB_TXT_PMR_ALIGN;
+	      mld->min_addr = ALIGN_UP (mld->min_addr, GRUB_TXT_PMR_ALIGN);
+	      mld->align = GRUB_TXT_PMR_ALIGN;
+	    }
```
krystian-hebel:
```suggestion
	  if (mld->align < GRUB_TXT_PMR_ALIGN)
	    mld->align = GRUB_TXT_PMR_ALIGN;

	  mld->min_addr = (mld->min_addr > slparams->mle_ptab_size) ?
			  (mld->min_addr - slparams->mle_ptab_size) : mld->align;
	  mld->min_addr = ALIGN_UP (mld->min_addr, mld->align);
```

grub-core/loader/multiboot_elfxx.c
krystian-hebel: I wrote it the other way around, this would break if `mld->align` is more restrictive than size of PTAB, e.g. when MB2 header asks for 2MB alignment (and this is where physical target points to) and `mle_ptab_size` is anything that isn't multiply of 2MB. This would be the case for Xen, but its relocatable tag is optional and it can handle being loaded at other addresses, but in general there could be a problem.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -180,6 +259,26 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
         }
     }
 
+  if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+    {
+      slparams->mle_header_offset = 0xffffffff;
+
+      for (mle_hdr_offset = 0;  mle_hdr_offset < 0x1000; mle_hdr_offset += 16)
+	{
+	  if ( !grub_memcmp ((void *)(slparams->mle_start + mle_hdr_offset), GRUB_TXT_MLE_UUID, 16) )
```
krystian-hebel: `slparams->mle_start` is `physical_target_address`, you shouldn't assume it is already populated. You must check against `virtual_current_address`, i.e. at `source`, even if in this case they are the same.

grub-core/loader/multiboot_elfxx.c
miczyg1: Then how can we trust what was passed here? I.e. the kernel and initrd and other params? IIRC the Xen cmdline is passed via MBI2 too, so it can't be ignored.

krystian-hebel: Xen should measure it to different PCR (to isolate code from configuration) before use, just as Linux does (or should do, it lacks "before use" part) with zero page. There is another task created to develop this.

I am aware that this will be different from approach taken on AMD platforms, but since we can't change how SINIT ACM works, our options are limited.

grub-core/loader/multiboot_elfxx.c
miczyg1: mle_ptab_size is at leas t2MB in size as I align it up to PMR/2MB

grub-core/loader/multiboot_elfxx.c
miczyg1: Fixed

grub-core/loader/multiboot_mbi2.c
miczyg1: Changed how MLE header is retrieved due to bugs how I searched it. So this change is removed

grub-core/loader/multiboot_mbi2.c
miczyg1: Outdated.

grub-core/loader/multiboot_mbi2.c
miczyg1: Outdated

grub-core/loader/multiboot_mbi2.c
miczyg1: Outdated.

grub-core/loader/multiboot_mbi2.c
miczyg1: Outdated.

grub-core/loader/i386/txt/txt.c
miczyg1: Yes, it was a mistake here. Fixed

grub-core/loader/multiboot_mbi2.c
miczyg1: Ad .1. they have to (at least this is what Linux did).
Ad. 2. yes
Ad. 3. Yes
Ad. 4. Yes

> Then MLE is set to everything between 2-4, including padding at the end, and that is measured to PCR17 by ACM, right?

Yes

grub-core/loader/multiboot_elfxx.c
miczyg1: Okay so I will need to rewrite a bit of it... Anyways I need to keep a record of where MBI is...

grub-core/loader/multiboot_elfxx.c
miczyg1: `slparams->mle_start` is already populated ~100 lines of code before. Secondly, `source` may change when loadable segments are loaded to memory, so I can't rely on that.

grub-core/loader/i386/txt/txt.c
miczyg1: FIxed

grub-core/loader/multiboot_elfxx.c
miczyg1: Outdated.

grub-core/loader/multiboot_mbi2.c
miczyg1: Outdated.

grub-core/loader/multiboot_elfxx.c
krystian-hebel: The address is populated, but if data needs to be relocated, it is done by relocator just before handing off execution to payload. You can also move these lines closer to where `source` is saved.

grub-core/loader/multiboot_mbi2.c
miczyg1: Fixed

grub-core/loader/multiboot_elfxx.c
miczyg1: So are you going to measure MBI in the assembly in Xen?

krystian-hebel: If you did review https://github.com/3mdeb/xen/pull/11 you would know that it can be done in C :slightly_smiling_face:

grub-core/loader/multiboot_mbi2.c
miczyg1: Outdated

grub-core/loader/multiboot_elfxx.c
miczyg1: Linux implementation has the mle_size aligned to PMR align (2MB), so I will simply leave load_size aligned up to 2MBs here without MBI.

grub-core/loader/multiboot_elfxx.c
miczyg1: Changed to source

grub-core/loader/multiboot_elfxx.c
miczyg1: Now MBI is not part of MLE

grub-core/loader/multiboot_mbi2.c
miczyg1: Remvoed. Now saving the target MBI physical address into boot_params_addr. THe MBI allocation is done as previously

grub-core/loader/multiboot_elfxx.c
miczyg1: Separated the allocation of MLE page tables from the multiboot kernel. The checks for alignment already ensure that there will be size for page tables below the allocated buffer for the kernel. Second allocation takes place when the load_base_addr is know and it serves as the upper boundary of the allocation (mle_ptab_size subtracted of course). This solves the `mld->align` problem I believe

grub-core/loader/multiboot.c
miczyg1: No idea how to do that better for now, multiboot is scattered around multiple files...

grub-core/loader/multiboot_elfxx.c
```diff
@@ -101,30 +106,96 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+	{
+#ifndef GRUB_USE_MULTIBOOT2
+	  return grub_error (GRUB_ERR_BAD_OS, "Only multiboot2 supported for slaunch");
+#else
+	  slparams->mle_start = mld->load_base_addr;
+
+          err = grub_relocator_alloc_chunk_align (GRUB_MULTIBOOT (relocator), &ch,
+					      0, mld->load_base_addr - slparams->mle_ptab_size,
+					      slparams->mle_ptab_size, GRUB_TXT_PMR_ALIGN,
+					      GRUB_RELOCATOR_PREFERENCE_NONE, 1);
```
krystian-hebel: Inconsistent indentation (spaces instead of tabs, lines not aligned to bracket).

grub-core/loader/multiboot_elfxx.c
krystian-hebel: By having MLE size aligned to 2MB we're wasting time measuring empty space. In fact, I don't see why this have to be aligned since PMR covers whole memory and I haven't found anything in TXT docs that would require such alignment.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -101,30 +106,96 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+	  slparams->mle_ptab_size = grub_txt_get_mle_ptab_size (total_size);
+	  slparams->mle_ptab_size = ALIGN_UP (slparams->mle_ptab_size, GRUB_TXT_PMR_ALIGN);
+
+	  /* Do not go below GRUB_TXT_PMR_ALIGN. */
+	  if (mld->align < GRUB_TXT_PMR_ALIGN)
+	    mld->align = GRUB_TXT_PMR_ALIGN;
+
+	  mld->min_addr = (mld->min_addr > slparams->mle_ptab_size) ?
+			  (mld->min_addr - slparams->mle_ptab_size) : mld->align;
+	  mld->min_addr = ALIGN_UP (mld->min_addr, mld->align);
```
krystian-hebel: I don't think this line is necessary since `mld->align` is taken into account by `grub_relocator_alloc_chunk_align()`.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -101,30 +106,96 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+
+	  slparams->mle_ptab_size = grub_txt_get_mle_ptab_size (total_size);
+	  slparams->mle_ptab_size = ALIGN_UP (slparams->mle_ptab_size, GRUB_TXT_PMR_ALIGN);
+
+	  /* Do not go below GRUB_TXT_PMR_ALIGN. */
+	  if (mld->align < GRUB_TXT_PMR_ALIGN)
+	    mld->align = GRUB_TXT_PMR_ALIGN;
+
+	  mld->min_addr = (mld->min_addr > slparams->mle_ptab_size) ?
+			  (mld->min_addr - slparams->mle_ptab_size) : mld->align;
```
krystian-hebel: This line moves min_addr lower, why?

grub-core/loader/multiboot_elfxx.c
miczyg1: This is what Linux implementation does, I have simply replicated same approach.

grub-core/loader/multiboot_elfxx.c
miczyg1: It is, because we change min_addr by subtracting slparams->mle_ptab_size which does not have to be `mld->align`'ed. BTW should it be ALIGN_DOWN?

grub-core/loader/multiboot_elfxx.c
miczyg1: To avoid situations where there is no space for page tables below the ELF image

krystian-hebel: But page tables are allocated later, they are not included in this chunk as far as I can tell.

miczyg1: Yes, but does that mean we cannot ensure there will be space for page tables here? What is the point if we will fail with page tables allocation later 100% of times for given binary which will always get allocated below 2MB?

krystian-hebel: Then you should add to `min_addr`, not subtract from it.

grub-core/loader/multiboot_elfxx.c
krystian-hebel: OK, until we get some confirmation as to why it is done this way, we can leave it as it is.

I still can't get expected MLE measurements, when compared with (uncompressed) file contents, no matter which alignment is used. Padding should be cleared to 0, but can't tell for sure. I don't feel like dumping 4MB of memory through serial to compare it, but I may have to... Another option is that there is something wrong with page tables and something else is measured, but I doubt it, after all it jumps into proper place in Xen later, which is also based on the same page tables.

miczyg1: Just uncompressing isn't enough. Do you strip the ELF header? Do you calculate measurement based on the loadable section only?

grub-core/loader/multiboot_elfxx.c
miczyg1: Removed it completely. Still boots

grub-core/loader/multiboot_elfxx.c
miczyg1: Removed it completely and still bots

grub-core/loader/multiboot_elfxx.c
miczyg1: Fixed

grub-core/loader/multiboot_elfxx.c
krystian-hebel: Obviously. I use something like this:

```
$ objdump xen -p

xen:     file format elf32-i386

Program Header:
    LOAD off    0x00000080 vaddr 0x00200000 paddr 0x00200000 align 2**6
         filesz 0x002c2900 memsz 0x003f8000 flags rwx
    NOTE off    0x00206858 vaddr 0x002067d8 paddr 0x002067d8 align 2**2
         filesz 0x00000024 memsz 0x00000024 flags r--

$ dd if=xen ibs=1 skip=$((0x80)) count=$((0x2c2900)) obs=$((0x3f8000)) conv=sync,notrunc | dd ibs=$((0x3f8000)) conv=sync,notrunc | sha1sum
```

This is Xen with additional debug lines added so your sizes will probably be different.

miczyg1: Anyway no idea what is going on.

krystian-hebel: OK, I'm marking it as resolved, since I don't see anything wrong with the way it is configured. Aligning everything to PMR may be an overkill, but it works. We may get back to it if/when something breaks for newer platforms, e.g. Daniel mentioned that page tables have to live directly below MLE, with no padding, which isn't true for us at the moment.

krystian-hebel (approved):

PR #9, opened 2023-07-29
Add GitHub Actions

SergiiDmytruk:
Automatically build `intal-txt-aem*` branches and pull requests by making a list of patches, putting them into `qubes-src/vmm-xen/`, updating spec-file and building the component to get RPM packages.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,18 @@
+name: Test build and package QubesOS RPMs
+
+on:
+  push:
+    branches:
+      - 'intel-txt-aem*'
+  pull_request_target:
+    branches: # this is to avoid running the workflow twice in a PR
+      - 'intel-txt-aem*'
+
+jobs:
+  qubes-dom0-package:
+    uses: TrenchBoot/.github/.github/workflows/qubes-dom0-package.yml@master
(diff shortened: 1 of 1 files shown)


SergiiDmytruk:
This fails because I moved from `intel-txt-aem-2.06` to `intel-txt-aem` and now patches don't apply.  I tried using Qubes package v2.04-3, but then it fails with:
```
No matching package to install: 'libusb-devel'
```
Should `intel-txt-aem` branch be updated to be based on grub 2.06?

miczyg1:
> Should intel-txt-aem branch be updated to be based on grub 2.06?

We should simply rely on `intel-txt-aem-2.06` which is based on v2.06. Qubes OS should also upgrade to 2.06 from 2.04

SergiiDmytruk:
Qubes OS is already using 2.06.   So this changes should be based on `intel-txt-aem-2.06` and the PR should target `intel-txt-aem-2.06`?

miczyg1:
> Qubes OS is already using 2.06. So this changes should be based on `intel-txt-aem-2.06` and the PR should target `intel-txt-aem-2.06`?

Exactly

macpijan (approved):

PR #10, opened 2023-08-18
Intel txt aem 2.06 new acm fixes

krystian-hebel:


grub-core/loader/i386/txt/txt.c
```diff
@@ -640,8 +640,7 @@ init_txt_heap (struct grub_slaunch_params *slparams, struct grub_txt_acm_header
 
   grub_dprintf ("slaunch", "SINIT capabilities %08x\n", sinit_caps);
 
-  os_sinit_data->capabilities = GRUB_TXT_CAPS_TPM_12_NO_LEGACY_PCR_USAGE |
-				GRUB_TXT_CAPS_TPM_12_AUTH_PCR_USAGE;
+  os_sinit_data->capabilities = GRUB_TXT_CAPS_TPM_12_AUTH_PCR_USAGE;
```
SergiiDmytruk, line 643: "No longer supported" in the guide doesn't seem to suggest the code needs to change in any way.  Sounds more like values of both bits aren't even checked by new ACMs, but might still be taken into account by older ones.  So I'm wondering whether this can break something.

miczyg1: AUTH_PCR_USAGE will take precedence over LEGACY_PCR_USAGE if both are requetsed/supported so this should still work as before.

Moreover it will work on older platforms too, which do not support AUTH_PCR_USAGE, because when NO_LEGACY_PCR_USAGE is not set (this bit is in negative logic, meaning 0 is supported), the legacy PCRs will be extended

miczyg1 (approved):
@SergiiDmytruk If you agree to my comment, please resolve it so we can merge. You can check out my reasoning in https://github.com/TrenchBoot/documentation/blob/master/references/intel-txt-software-development-guide.pdf
(Table 2-2)

SergiiDmytruk (approved):

PR #13, opened 2023-10-10
Import SLRT support for secure launch

SergiiDmytruk:
This addresses https://github.com/TrenchBoot/trenchboot-issues/issues/17

Commits were cherry-picked according to https://github.com/TrenchBoot/trenchboot-issues/issues/17#issuecomment-1494679529 with minor modifications where necessary (bugs, typos, compilation warnings/errors).  `grub-sl-fc-38-dlstub` contains several more patches related to code reorganization and EFI, _these were not included_.

The changes weren't tested.

### DIFF
--- a/grub-core/loader/i386/slaunch.c
+++ b/grub-core/loader/i386/slaunch.c
@@ -23,11 +23,13 @@
 #include <grub/misc.h>
 #include <grub/types.h>
 #include <grub/dl.h>
+#include <grub/slr_table.h>
 #include <grub/cpu/relocator.h>
 #include <grub/i386/cpuid.h>
 #include <grub/i386/msr.h>
 #include <grub/i386/mmio.h>
 #include <grub/i386/slaunch.h>
+#include <grub/i386/tpm.h>
 #include <grub/i386/txt.h>
 
 GRUB_MOD_LICENSE ("GPLv3+");
@@ -37,6 +39,13 @@ static grub_uint32_t slp = SLP_NONE;
 static void *slaunch_module = NULL;
 
 static struct grub_slaunch_params slparams;
+
+/* Area to collect and build SLR Table information. */
+static struct grub_slr_entry_dl_info slr_dl_info_staging;
+static struct grub_slr_entry_log_info slr_log_info_staging;
+static grub_uint8_t slr_policy_buf[GRUB_PAGE_SIZE];
+static struct grub_slr_entry_policy *slr_policy_staging =
+    (struct grub_slr_entry_policy *)slr_policy_buf;
 
 grub_uint32_t
 grub_slaunch_platform_type (void)
@@ -54,6 +63,84 @@ struct grub_slaunch_params *
 grub_slaunch_params (void)
 {
   return &slparams;
+}
+
+void
+grub_slaunch_init_slrt_storage (int arch)
+{
+  struct grub_txt_mle_header *mle_header =
+      (void *) ((grub_uint8_t *) slparams.mle_mem + slparams.mle_header_offset);
+
+  /* Setup the generic bits of the SLRT. */
+  grub_slr_init_table(slparams.slr_table_mem, arch, slparams.slr_table_size);
+
+  /* Setup DCE and DLME information. */
+  slr_dl_info_staging.hdr.tag = GRUB_SLR_ENTRY_DL_INFO;
+  slr_dl_info_staging.hdr.size = sizeof(struct grub_slr_entry_dl_info);
+  slr_dl_info_staging.dce_base = slparams.dce_base;
+  slr_dl_info_staging.dce_size = slparams.dce_size;
+  slr_dl_info_staging.dlme_entry = mle_header->entry_point;
+
+  slr_log_info_staging.hdr.tag = GRUB_SLR_ENTRY_LOG_INFO;
+  slr_log_info_staging.hdr.size = sizeof(struct grub_slr_entry_log_info);
+  slr_log_info_staging.addr = slparams.tpm_evt_log_base;
+  slr_log_info_staging.size = slparams.tpm_evt_log_size;
+  slr_log_info_staging.format =
+        (grub_get_tpm_ver () == GRUB_TPM_20) ?
+        GRUB_SLR_DRTM_TPM20_LOG : GRUB_SLR_DRTM_TPM12_LOG;
+
+  slr_policy_staging->hdr.tag = GRUB_SLR_ENTRY_DRTM_POLICY;
+  slr_policy_staging->hdr.size = sizeof(struct grub_slr_entry_policy);
+  slr_policy_staging->revision = GRUB_SLR_TABLE_REVISION;
+  slr_policy_staging->nr_entries = 0;
+
+  /* The SLR table should be measured too, at least parts of it. */
+  grub_slaunch_add_slrt_policy_entry (18,
+                                      GRUB_SLR_ET_SLRT,
+                                      GRUB_SLR_POLICY_IMPLICIT_SIZE,
+                                      slparams.slr_table_base,
+                                      /*size=*/0,
+                                      "Measured SLR Table");
+}
+
+void
+grub_slaunch_add_slrt_policy_entry (grub_uint16_t pcr,
+                                    grub_uint16_t entity_type,
+                                    grub_uint16_t flags,
+                                    grub_uint64_t entity,
+                                    grub_uint64_t size,
+                                    const char *evt_info)
+{
+  struct grub_slr_policy_entry *entry =
+    (void *)((grub_uint8_t *)slr_policy_staging +
+             sizeof(struct grub_slr_entry_policy) +
+             slr_policy_staging->nr_entries*sizeof(*entry));
+
+  if (slr_policy_staging->hdr.size + sizeof(*entry) > sizeof(slr_policy_buf))
+      grub_fatal("Not enough space for adding policy entry: %s!  The buffer is full.",
+                 evt_info);
+
+  entry->pcr = pcr;
+  entry->entity_type = entity_type;
+  entry->flags = flags;
+  entry->entity = entity;
+  entry->size = size;
+
+  grub_strncpy(entry->evt_info, evt_info, sizeof(entry->evt_info) - 1);
+  entry->evt_info[sizeof(entry->evt_info) - 1] = '\0';
+
+  slr_policy_staging->hdr.size += sizeof(*entry);
+  ++slr_policy_staging->nr_entries;
+}
+
+void
+grub_slaunch_finish_slr_table (void)
+{
+  struct grub_slr_table *slr_table = slparams.slr_table_mem;
+
+  grub_slr_add_entry (slr_table, &slr_dl_info_staging.hdr);
+  grub_slr_add_entry (slr_table, &slr_log_info_staging.hdr);
+  grub_slr_add_entry (slr_table, &slr_policy_staging->hdr);
 }
 
 static grub_err_t
--- a/grub-core/loader/i386/txt/txt.c
+++ b/grub-core/loader/i386/txt/txt.c
@@ -59,6 +59,7 @@
 #include <grub/types.h>
 #include <grub/dl.h>
 #include <grub/acpi.h>
+#include <grub/slr_table.h>
 #include <grub/cpu/relocator.h>
 #include <grub/i386/cpuid.h>
 #include <grub/i386/msr.h>
@@ -73,6 +74,8 @@
 #define OS_SINIT_DATA_TPM_20_VER	7
 
 #define OS_SINIT_DATA_MIN_VER		OS_SINIT_DATA_TPM_12_VER
+
+static struct grub_slr_entry_intel_info slr_intel_info_staging = {0};
 
 static grub_err_t
 enable_smx_mode (void)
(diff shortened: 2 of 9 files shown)


grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +560,138 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
...
+}
+
+static void
+setup_slrt_policy (struct grub_slaunch_params *slparams,
+                   struct grub_txt_os_mle_data *os_mle_data)
+{
+  struct grub_slr_policy_entry *entry =
+    (struct grub_slr_policy_entry *)((grub_uint8_t *)slr_policy_staging +
+                                     sizeof(struct grub_slr_entry_policy));
+  struct linux_kernel_params *boot_params = (void *) (grub_addr_t) slparams->boot_params_addr;
```
krystian-hebel: This assumes that only Linux will be ever set up through `setup_slrt_policy()`. I guess the best course is to extract common parts (SLRT, OS2MLE) and from `{linux,multiboot}.c` call `{linux,multiboot2}_setup_slrt_policy()`, which in turn calls the function that fills common part.

grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +560,138 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
...
+  if (boot_params->setup_data)
+    {
+      entry->pcr = 18;
+      entry->entity_type = GRUB_SLR_ET_SETUP_DATA;
+      entry->entity = boot_params->setup_data;
+      entry->flags |= GRUB_SLR_POLICY_IMPLICIT_SIZE;
+      grub_strcpy (&entry->evt_info[0], "Measured Kernel setup_data");
+    }
+  else
+      entry->entity_type = GRUB_SLR_ET_UNUSED;
```
krystian-hebel: Spec update: explicitly mention that such gaps may exist in policy table, preferably with explanation when and why.

grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +560,138 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
...
+  entry++;
+
+  entry->pcr = 18;
+  entry->entity_type = GRUB_SLR_ET_CMDLINE;
+  /* TODO the cmdline ptr can have hi bits but for now assume always < 32G */
```
krystian-hebel:
```suggestion
  /* TODO the cmdline ptr can have hi bits but for now assume always < 4G */
```

grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +560,138 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
...
+  entry->entity = boot_params->cmd_line_ptr;
+  entry->size = boot_params->cmdline_size;
+  grub_strcpy (&entry->evt_info[0], "Measured Kernel command line");
+  entry++;
+
+  if (!grub_memcmp(&efi_info->efi_signature, "EL64", sizeof(grub_uint32_t)))
```
krystian-hebel: We probably should check if `efi_info != NULL` before this, which may indicate legacy boot.

grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +560,138 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
...
+
+static void
+setup_slr_table (struct grub_slaunch_params *slparams)
+{
+  struct grub_slr_table *slr_table = (void *) (grub_addr_t) slparams->slr_table_base;
+
+  grub_slr_add_entry (slr_table, &slr_dl_info_staging.hdr);
+  grub_slr_add_entry (slr_table, &slr_log_info_staging.hdr);
+  grub_slr_add_entry (slr_table, &slr_policy_staging->hdr);
+  grub_slr_add_entry (slr_table, &slr_intel_info_staging.hdr);
```
krystian-hebel, line 554: Architecture specific. Should be easy to check the architecture with `grub_slaunch_platform_type()`. We could also remove this one entry from here and call it directly from `grub_txt_boot_prepare()`, or even `set_txt_info_ptr()` after renaming than function.

grub-core/loader/i386/txt/txt.c
```diff
@@ -59,6 +59,7 @@
 #include <grub/types.h>
 #include <grub/dl.h>
 #include <grub/acpi.h>
+#include <grub/slr_table.h>
```
krystian-hebel, line 62: Applies to all of the SLRT stuff: this isn't specific to TXT. Most of the functions and structures should be in `slaunch.{c,h}` files instead.

grub-core/loader/i386/txt/txt.c
```diff
@@ -74,6 +75,30 @@
 
 #define OS_SINIT_DATA_MIN_VER		OS_SINIT_DATA_TPM_12_VER
 
+#define SLR_CURRENT_REVISION		1
+#define SLR_MAX_POLICY_ENTRIES		7
+
+struct grub_efi_info
```
krystian-hebel: Is this the format used by Linux? If yes, the structure name should probably reflect that.

include/grub/slr_table.h
```diff
@@ -0,0 +1,296 @@
...
+#define GRUB_SLR_ENTRY_DRTM_POLICY	0x0003
+#define GRUB_SLR_ENTRY_INTEL_INFO	0x0004
+#define GRUB_SLR_ENTRY_AMD_INFO		0x0005
+#define GRUB_SLR_ENTRY_ARM_INFO		0x0006
+#define GRUB_SLR_ENTRY_UEFI_INFO	0x0007
+#define GRUB_SLR_ENTRY_UEFI_CONFIG	0x0008
+#define GRUB_SLR_ENTRY_END		0xffff
+
+/* Entity Types */
+#define GRUB_SLR_ET_UNSPECIFIED		0x0000
```
krystian-hebel, line 68: Spec update: according to the specification, `SLR_ET_UNSPECIFIED` can be used to include measurements of generic memory regions. However, this value is also the initial value after SLRT buffer is cleared. `nr_entries` in the header specifies total number of entries available and not the number of actually used ones, so there is no easy way of knowing when to stop parsing policy entries. I suggest we rename type 0 to mark end of table and assign different number to `SLR_ET_UNSPECIFIED`.

include/grub/slr_table.h
```diff
@@ -0,0 +1,296 @@
...
+{
+  grub_uint16_t pcr;
+  grub_uint16_t reserved;
+  grub_uint64_t cfg; /* address or value */
+  grub_uint32_t size;
+  char evt_info[GRUB_TPM_EVENT_INFO_LENGTH];
+} GRUB_PACKED;
+
+static inline void *
+grub_slr_end_of_entrys (struct grub_slr_table *table)
```
krystian-hebel:
```suggestion
grub_slr_end_of_entries (struct grub_slr_table *table)
```

krystian-hebel (commented):
I've left some comments. Most of them are hints for the next steps in separate PRs, but there are some typos and NULL check that should probably be addressed here.

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: I was wondering where 32G came from.

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: Given that it's computed as
```c
  efi_info = (struct grub_efi_info *)((grub_uint8_t *)&(boot_params->v0208)
                                       + 2*sizeof(grub_uint32_t));
```
it can't be `NULL`.  The check for legacy boot would have to be different.

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: Yes, it's part of `struct linux_kernel_params`.  Renamed to `linux_params_efi_info`.

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: Moved to `set_txt_info_ptr()` after renaming it to `setup_txt_slrt_entry()`.

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: See d6d6a8def

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: See 564309737

SergiiDmytruk (commented):
I didn't squash commits yet so it's easier to see differences.  See https://github.com/TrenchBoot/grub/compare/366d53cf8f4cb9af5aa01508ded0564b5357f70d..f18f1e3be for changes in top commit and d6d6a8def with 564309737 for larger changes.

There are no more gaps or fixed size of policy entries.  Things are added to the list dynamically from different places.  This required making callers of `grub_txt_boot_prepare()` responsible for finishing building SLRT.

grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +538,19 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
   elog->next_event_offset = sizeof(*elog);
 }
 
+static void
+setup_txt_slrt_entry (struct grub_slaunch_params *slparams,
+                      struct grub_txt_os_mle_data *os_mle_data)
+{
+  struct grub_slr_table *slr_table = (void *) (grub_addr_t) slparams->slr_table_base;
```
krystian-hebel:
```suggestion
  struct grub_slr_table *slr_table = (void *) (grub_addr_t) slparams->slr_table_mem;
```

grub-core/loader/i386/slaunch.c
```diff
@@ -56,6 +65,85 @@ grub_slaunch_params (void)
...
+  entry->evt_info[sizeof(entry->evt_info) - 1] = '\0';
+
+  slr_policy_staging->hdr.size += sizeof(*entry);
+  ++slr_policy_staging->nr_entries;
+}
+
+void
+grub_slaunch_finish_slr_table (void)
+{
+  struct grub_slr_table *slr_table = (void *) (grub_addr_t) slparams.slr_table_base;
```
krystian-hebel: In most cases they are the same, but not always, and it's hard to debug cases when they aren't.

```suggestion
  struct grub_slr_table *slr_table = (void *) (grub_addr_t) slparams.slr_table_mem;
```

grub-core/loader/i386/slaunch.c
```diff
@@ -56,6 +65,85 @@ grub_slaunch_params (void)
   return &slparams;
 }
 
+void
+grub_slaunch_init_slrt_storage (int arch)
+{
+  struct grub_txt_mle_header *mle_header;
+
+  mle_header = (struct grub_txt_mle_header *)(grub_addr_t) (slparams.mle_start + slparams.mle_header_offset);
```
krystian-hebel: This also suffers from potential current vs target address mismatch. While it should be possible to calculate this using `mle_ptab_mem` and `mle_ptab_size` for Linux, Multiboot2 path takes another approach where it wouldn't work. Adding separate field to `struct grub_slaunch_params` for current address of `mle_start` would also make it easier to follow.

grub-core/loader/i386/txt/txt.c
```diff
@@ -535,6 +538,19 @@ grub_txt_init_tpm_event_log (void *buf, grub_size_t size)
...
+  struct grub_slr_entry_hdr *txt_info;
+
+  grub_slr_add_entry (slr_table, &slr_intel_info_staging.hdr);
+
+  txt_info = grub_slr_next_entry_by_tag (slr_table, NULL, GRUB_SLR_ENTRY_INTEL_INFO);
+  os_mle_data->txt_info = (grub_addr_t) txt_info;
```
krystian-hebel: Here on the other hand we need target address. Something like `slparams->slr_table_base + (txt_info - slparams->slr_table_mem)` with appropriate casts as needed.

grub-core/loader/i386/txt/txt.c
```diff
@@ -567,16 +584,22 @@ init_txt_heap (struct grub_slaunch_params *slparams, struct grub_txt_acm_header
...
   os_mle_data->ap_wake_block = slparams->ap_wake_block;
   os_mle_data->ap_wake_block_size = slparams->ap_wake_block_size;
 
-  os_mle_data->evtlog_addr = slparams->tpm_evt_log_base;
-  os_mle_data->evtlog_size = slparams->tpm_evt_log_size;
-
+  /* Save the BSPs MTRR state so post launch can restore it. */
   grub_dprintf ("slaunch", "Saving MTRRs to OS MLE data\n");
-  save_mtrrs (os_mle_data);
+  save_mtrrs (&saved_mtrrs_state);
```
krystian-hebel: Can we do `save_mtrrs (&(slr_intel_info_staging.saved_bsp_mtrrs));` instead? This would save us from having to `memcpy` it later, as well as some stack usage, however it would probably require `save_mtrrs()` to zero all MTRRs between `mtrr_vcnt` and `GRUB_TXT_VARIABLE_MTRRS_LENGTH`.

grub-core/loader/i386/txt/txt.c
krystian-hebel: No longer applies, no gaps are created anymore.

krystian-hebel (commented):
Most of the comments are for different addresses before and after relocator, some of them were already present in the previous phase.

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-e86942d6c01510d4eb71b81ebde4ba5e2f75f6dd8c4e4a5b1274f527b2793579L545

grub-core/loader/i386/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-33a2d35c8932ea248b7465a187fc4603651840c19cb98630f3d5276d9163d57bL140

grub-core/loader/i386/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-33a2d35c8932ea248b7465a187fc4603651840c19cb98630f3d5276d9163d57bL71
Defined here:
https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-7454bb1422e946af15dee15e8ed5e222430f7652039c50757eb6c96ef1af1faeR43
Set here:
https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-8c69f5f8f99bf5f4195720ea867e6e8fe273f9261ce151f58222e0d73b34c208R264
Also updated this usage:
https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-e86942d6c01510d4eb71b81ebde4ba5e2f75f6dd8c4e4a5b1274f527b2793579R983

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-e86942d6c01510d4eb71b81ebde4ba5e2f75f6dd8c4e4a5b1274f527b2793579R557

grub-core/loader/i386/txt/txt.c
SergiiDmytruk: Did that:
https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-e86942d6c01510d4eb71b81ebde4ba5e2f75f6dd8c4e4a5b1274f527b2793579L568
And zeroed out those entries:
https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-e86942d6c01510d4eb71b81ebde4ba5e2f75f6dd8c4e4a5b1274f527b2793579R277
Removed original structures use of which was non-obvious with `memcpy()`:
https://github.com/TrenchBoot/grub/compare/564309737c27fd2378e17a813ab0391b9ac82ead..d87efc08c3861127e655c68bde2117642fa827da#diff-e80e689d00fce50be9d705cff6a6fb841a2be25a6a52fc57f59fe3fc5c0d4227L356

krystian-hebel (approved):

include/grub/slr_table.h
krystian-hebel: Never mind, I misread the specification, this field is indeed the current number of entries in the table.

PR #14, opened 2023-10-25
multiboot2: Switch to using SLRT interface

SergiiDmytruk:


grub-core/loader/multiboot_mbi2.c
```diff
@@ -734,8 +735,77 @@ retrieve_video_parameters (grub_properly_aligned_t **ptrorig)
   return GRUB_ERR_NONE;
 }
 
+void
+grub_multiboot2_setup_slr_table (grub_uint32_t mbi_target, void *mbi_mem)
```
krystian-hebel: Whole MBI is measured by Xen, so there is little reason to measure parts of it again. We probably should just add an entry for MBI (either with implicit size, as it currently is read from MBI itself, or explicitly set in SLRT entry for extra security) and each module, e.g. at the time they are iterated over in `grub_multiboot2_make_mbi()`. In that case, TXT stuff would probably have to be moved from `normal_boot()` to `grub_multiboot_boot()`, which would generate some problems with UEFI path, but we don't care about it (yet) so throwing an error would be enough for now.

include/grub/slr_table.h
```diff
@@ -73,6 +73,8 @@
 #define GRUB_SLR_ET_UEFI_MEMMAP		0x0005
 #define GRUB_SLR_ET_RAMDISK		0x0006
 #define GRUB_SLR_ET_TXT_OS2MLE		0x0010
+#define GRUB_SLR_ET_MODULE		0x0011
```
krystian-hebel: I think the idea was to group entry types, and this would fit in the same group as earlier types (0x0000-0x000f). By the look of it, 0x001? is used for things specific to TXT (or vendor-specific in general).

---

This applies to existing entry types as well: I think the names aren't descriptive enough. GRUB can boot many different operating systems, and specification is supposedly bootloader-agnostic. `GRUB_SLR_ET_SETUP_DATA` by itself doesn't tell too much. IMHO we should consider renaming it to `GRUB_SLR_ET_LINUX_SETUP_DATA`, even though it makes the name longer, similarly for `BOOT_PARAMS` and `MODULE`. This is however an issue with the specification, so I don't expect an immediate action here (except for numbering mentioned above).

grub-core/loader/multiboot_mbi2.c
SergiiDmytruk: Can't MBI change while staying the same, say, when pointer values change because allocation code was updated?  Or is a failure in this case expected?

> e.g. at the time they are iterated over in `grub_multiboot2_make_mbi()`

If you want to avoid parsing of MBI, can just iterate over `modules` global variable which shouldn't be freed.

include/grub/slr_table.h
SergiiDmytruk: Well, I noticed that these are quite specific and deliberately made the name generic :)  This also helped to avoid deciding whether name should mention Multiboot version.  And that gap without any explanation looked like a deliberate one (e.g., some things were removed).

grub-core/loader/multiboot_mbi2.c
krystian-hebel: > Can't MBI change while staying the same, say, when pointer values change because allocation code was updated?

Fair point, I haven't thought about it. This would require strict description of what isn't measured, and this would probably change with new tags added in the future. On the other hand, such change would most likely make EFI_MMAP different as well, and it is measured.

krystian-hebel: Now that I think about it, SLRT was created by an entity that isn't measured in DRTM chain, so we can't trust it to be valid _until SLRT is measured_. Since addresses aren't measured according to Annex A, there is no guaranty that these addresses are the same as in MBI, so we may end up measuring different region that that which will be later executed. We would need to explicitly check in Xen if those addresses are the same. I wonder if Linux does the same for ramdisk, @dpsmith @rossphilipson any insights?

---

Going back on track, I'd say that MBI should be treated the same way Linux boot params are handled. It serves similar role, also contains pointers, and IIUC it is measured as a whole. One significant difference is that MBI has the command line embedded, while Linux holds just the pointer.

SergiiDmytruk: Given absence of any active discussion here, should we just do it some way and change things later if it turns out to be inadequate?

> Going back on track, I'd say that MBI should be treated the same way Linux boot params are handled. It serves similar role, also contains pointers, and IIUC it is measured as a whole.

It's measured as a whole and then pointers are followed to measure at least some other parts as well (setup data, command-line, UEFI memmap).

include/grub/slr_table.h
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/87f124a95cc8acc45103d9de33df11e8240d1776..94c7f4f9d2f33d3bf503c6356f779261674cf351#diff-fa8593325ef57d162ad9d9472e2e6571ec4dba071acb7da6e07515bbe8658a63

grub-core/loader/multiboot_mbi2.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/87f124a95cc8acc45103d9de33df11e8240d1776..94c7f4f9d2f33d3bf503c6356f779261674cf351#diff-285279252527f5bc911013189ac09a0880271f60dc8f28f28fe68c2f5d517c74

Most other changes replace `mbi_mem` with `mbi_size`.

grub-core/loader/multiboot.c
```diff
@@ -162,23 +162,58 @@ efi_boot (struct grub_relocator *rel __attribute__ ((unused)),
...
+        }
 
+      err = grub_txt_boot_prepare (slparams);
       if (err != GRUB_ERR_NONE)
 	{
 	  grub_printf ("TXT boot preparation failed");
 	  return;
 	}
 
+      GRUB_MULTIBOOT (setup_slr_table) (state.MULTIBOOT_MBI_REGISTER, mbi_size);
```
krystian-hebel: Why not use `grub_multiboot (get_mbi_size)()` instead of adding argument to multiple functions?

Also, the whole `if (state.edi == SLP_INTEL_TXT)` block could be put in `#ifdef GRUB_USE_MULTIBOOT2`, that way `grub_multiboot_setup_slr_table()` wouldn't be necessary, and `grub_multiboot_init_slparams()` could be renamed to `grub_multiboot2_init_slparams()` and also surrounded by `#ifdef GRUB_USE_MULTIBOOT2`.

SergiiDmytruk: > Why not use `grub_multiboot (get_mbi_size)()` instead of adding argument to multiple functions?

`grub_multiboot2_make_mbi()` uses result of `grub_multiboot2_get_mbi_size()` for memory allocation but not when storing MBI size inside MBI, so I'm not sure the values match.

> Also, the whole `if (state.edi == SLP_INTEL_TXT)` block could be put in `#ifdef GRUB_USE_MULTIBOOT2`, that way `grub_multiboot_setup_slr_table()` wouldn't be necessary, and `grub_multiboot_init_slparams()` could be renamed to `grub_multiboot2_init_slparams()` and also surrounded by `#ifdef GRUB_USE_MULTIBOOT2`.

[Done](https://github.com/TrenchBoot/grub/compare/94c7f4f9d2f33d3bf503c6356f779261674cf351..d6cd47831c8cf81429eb99439028b74902ba6049#diff-9748f842c85ed9a9821039ab14b2e4b79227be72b28b6e38ca876932e7fa917e) and noticed that MBI wasn't measured first, so had to do ee8e036e7f45c14ae8ab9a1b3d7276f0524bd86a.

grub-core/loader/multiboot.c
```diff
@@ -210,11 +229,48 @@ grub_multiboot_boot (void)
 #endif
   state.MULTIBOOT_ENTRY_REGISTER = GRUB_MULTIBOOT (payload_eip);
 
-  err = GRUB_MULTIBOOT (make_mbi) (&state.MULTIBOOT_MBI_REGISTER);
+#ifdef GRUB_USE_MULTIBOOT2
+  if (state.edi == SLP_INTEL_TXT)
```
krystian-hebel: This is before `normal_boot()` so `state.edi` isn't set yet. Please either move `state.edi = grub_slaunch_platform_type ();` from `normal_boot()` here, or compare `grub_slaunch_platform_type()` directly.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -843,6 +844,13 @@ grub_multiboot2_make_mbi (grub_uint32_t *target)
 	grub_memcpy (tag->cmdline, cur->cmdline, cur->cmdline_size);
 	ptrorig += ALIGN_UP (tag->size, MULTIBOOT_TAG_ALIGN)
 	  / sizeof (grub_properly_aligned_t);
+
+        grub_slaunch_add_slrt_policy_entry (17,
```
krystian-hebel: This should be removed, modules are added later by `grub_multiboot2_add_slrt_policy_entries()`. Also, it would be called for non-slaunch boots here.

grub-core/loader/multiboot.c
```diff
@@ -210,11 +229,48 @@ grub_multiboot_boot (void)
 #endif
   state.MULTIBOOT_ENTRY_REGISTER = GRUB_MULTIBOOT (payload_eip);
 
-  err = GRUB_MULTIBOOT (make_mbi) (&state.MULTIBOOT_MBI_REGISTER);
+#ifdef GRUB_USE_MULTIBOOT2
```
krystian-hebel, line 208: `grub_txt_boot_prepare()` called below expects `slparams->boot_params_addr` to be valid, so please move this below `make_mbi()`, because this is where `boot_params_addr` get set.

krystian-hebel (commented):
With these changes I can boot into Xen, however it still doesn't fully work. Depending on whether I comment out `grub_multiboot2_set_video_mode()` (to get final lines printed on VGA) Xen either hangs somewhere (if not commented out) or reboots with TXT.ERRORCODE set to 0x8000000a (commented out).

grub-core/loader/multiboot.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/d6cd47831c8cf81429eb99439028b74902ba6049..03579a740622f9fb5002f666b8bd4a93e40696a9#diff-9748f842c85ed9a9821039ab14b2e4b79227be72b28b6e38ca876932e7fa917eR238

grub-core/loader/multiboot_mbi2.c
SergiiDmytruk: That was a transitive change which I forgot to drop.

https://github.com/TrenchBoot/grub/compare/d6cd47831c8cf81429eb99439028b74902ba6049..03579a740622f9fb5002f666b8bd4a93e40696a9#diff-285279252527f5bc911013189ac09a0880271f60dc8f28f28fe68c2f5d517c74

grub-core/loader/multiboot.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/d6cd47831c8cf81429eb99439028b74902ba6049..03579a740622f9fb5002f666b8bd4a93e40696a9#diff-9748f842c85ed9a9821039ab14b2e4b79227be72b28b6e38ca876932e7fa917e

SergiiDmytruk:
> Depending on whether I comment out `grub_multiboot2_set_video_mode()` (to get final lines printed on VGA) Xen either hangs somewhere (if not commented out) or reboots with TXT.ERRORCODE set to 0x8000000a (commented out).

Which commit of https://github.com/TrenchBoot/xen/pull/7 was used?  0x8000000a is that "catch all" error code which doesn't mean much.

I guess this also tests changes from https://github.com/TrenchBoot/grub/pull/13.

grub-core/loader/multiboot.c
```diff
@@ -162,25 +163,46 @@ efi_boot (struct grub_relocator *rel __attribute__ ((unused)),
 #endif
 
 #if defined (__i386__) || defined (__x86_64__)
+#ifdef GRUB_USE_MULTIBOOT2
+static grub_err_t
+grub_multiboot2_init_slparams (struct grub_slaunch_params *slparams)
```
krystian-hebel: Static functions don't have to begin with `grub_*`, this is required only for externally called functions. While it isn't forbidden, I think dropping it would make the code easier to follow for people familiar with GRUB code style.

krystian-hebel: Actually, does this function have to be defined here? Maybe `multiboot_mbi2.c` would be a better place.

grub-core/loader/multiboot.c
```diff
@@ -210,11 +233,43 @@ grub_multiboot_boot (void)
...
 
-  err = GRUB_MULTIBOOT (make_mbi) (&state.MULTIBOOT_MBI_REGISTER);
+  err = GRUB_MULTIBOOT (make_mbi) (&state.MULTIBOOT_MBI_REGISTER, &mbi_size);
 
   if (err)
     return err;
 
+#ifdef GRUB_USE_MULTIBOOT2
+  if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+    {
```
krystian-hebel, line 210: How about extracting the body to new function in `multiboot_mbi2.c`? That way some of the called functions could be made static there as well, keeping namespace clearer, e.g. `grub_multiboot_add_slrt_policy_entries()` wouldn't have to be declared for MB1 at all. This would probably be enough to remove `#include <grub/slr_table.h>` from here.

grub-core/loader/multiboot.c
```diff
@@ -162,25 +163,46 @@ efi_boot (struct grub_relocator *rel __attribute__ ((unused)),
...
+  struct grub_txt_os_mle_data *os_mle_data;
+  grub_addr_t slrt;
+  grub_err_t err;
+  grub_relocator_chunk_t ch;
+
+  txt_heap = grub_txt_get_heap ();
+  os_mle_data = grub_txt_os_mle_data_start (txt_heap);
+  slrt = ALIGN_UP ((grub_addr_t) os_mle_data + sizeof(*os_mle_data), 4);
+
+  err = grub_relocator_alloc_chunk_addr (GRUB_MULTIBOOT (relocator), &ch,
```
krystian-hebel: It doesn't work that way. TXT heap is not maintained by GRUB relocator, it is detected as reserved memory. As a result, GRUB takes the next available range for current "virtual" address and continues with hope that the target address will become available before relocation happens. This isn't the case so relocation fails with out of memory error.

Instead it has to be manually reserved on TXT heap and taken for granted. See how it is done in [init_txt_heap()](https://github.com/TrenchBoot/grub/blob/e1bdaf2ef7638f56b9c059639b91a9c36596bfc1/grub-core/loader/i386/txt/txt.c#L587).

Order of execution is crucial, mostly because of how fragmented SLRT and TXT heap building is. This is not the only possible implementation, feel free to propose something else, but I would do it like this:
- reserve some memory for SLRT with `grub_malloc()` or its derivatives, not with relocator
- save pointer to memory reserved above as `slparams->slr_table_mem`, point `slparams.slr_table_base` to TXT heap
- fill SLRT as needed, when needed
- in `init_txt_heap()` check if `slparams.slr_table_base` points to TXT heap, if it does, make OS-MLE section appropriately bigger (page size plus alignment); some of SLRT may not be filled in yet so don't trust `slparams->slr_table_size`
- continue filling SLRT in `slparams->slr_table_mem`
- copy `slparams->slr_table_mem` to `slparams->slr_table_base` as late as possible, but only if `slparams->slr_table_base` points to TXT heap - `grub_slaunch_finish_slr_table()` seems like a logical place to do so

Conditional sizing and copying based on location of `slparams->slr_table_base` is there mostly to not break Linux implementation. If SLRT grows or TXT heap size shrinks in the future it should make it easier to just move the allocation elsewhere.

grub-core/loader/multiboot.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/e1bdaf2ef7638f56b9c059639b91a9c36596bfc1..b293777b584771e0cc62d9b538b23d23b652441b#diff-9748f842c85ed9a9821039ab14b2e4b79227be72b28b6e38ca876932e7fa917e

grub-core/loader/multiboot.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/e1bdaf2ef7638f56b9c059639b91a9c36596bfc1..b293777b584771e0cc62d9b538b23d23b652441b#diff-285279252527f5bc911013189ac09a0880271f60dc8f28f28fe68c2f5d517c74

grub-core/loader/multiboot.c
SergiiDmytruk: The whole of https://github.com/TrenchBoot/grub/compare/e1bdaf2ef7638f56b9c059639b91a9c36596bfc1..b293777b584771e0cc62d9b538b23d23b652441b

> * save pointer to memory reserved above as `slparams->slr_table_mem`, point `slparams.slr_table_base` to TXT heap

I added a constant for this instead of using TXT heap address, so it's less cryptic.

> * some of SLRT may not be filled in yet so don't trust `slparams->slr_table_size`

`slparams->slr_table_size` is maximum size of SLRT, so it can be trusted.

> * copy `slparams->slr_table_mem` to `slparams->slr_table_base` as late as possible, but only if `slparams->slr_table_base` points to TXT heap - `grub_slaunch_finish_slr_table()` seems like a logical place to do so

Did it right after `grub_slaunch_finish_slr_table()` to keep related code close to each other (the constant from above is used in the same function).

krystian-hebel (approved):
Seems that GRUB side works, but Xen still triple-faults. I'm approving this PR but I'll hold with merging until I can confirm that only Xen needs changes.

PR #16, opened 2024-01-27
Intel txt aem 2.06 testing

krystian-hebel:


docs/grub.texi
```diff
@@ -6019,9 +6019,10 @@ corresponding to the filename.
...
 will carry this out. GRUB will also not perform any measurements until the
-tpm module is loaded. As such it is recommended that the tpm module be built
-into @file{core.img} in order to avoid a potential gap in measurement between
-@file{core.img} being loaded and the tpm module being loaded.
+tpm_verifier module is loaded. As such it is recommended that the tpm_verifier
+module be built into @file{core.img} in order to avoid a potential gap in
+measurement between @file{core.img} being loaded and the tpm_verifier module
+being loaded.
 
 Measured boot is currently only supported on EFI platforms.
```
DemiMarie, line 6027: Is this still true?

grub-core/lib/i386/relocator32.S
```diff
@@ -110,11 +112,17 @@ VARIABLE(grub_relocator32_edx)
...
+	je	LOCAL(intel_txt)
+
 	.byte	0xea
 VARIABLE(grub_relocator32_eip)
 	.long	0
 	.word	CODE_SEGMENT
 
+LOCAL(intel_txt):
+	getsec
+
```
DemiMarie, line 125: I’m not familiar with assembly language, but this seems wrong: if the `je` is not taken, execution will fall through to `getsec` anyway, with what is presumably an invalid parameter for that opcode.

grub-core/loader/i386/linux.c
```diff
@@ -150,11 +173,36 @@ allocate_pages (grub_size_t prot_size, grub_size_t *align,
...
-  prot_size = page_align (prot_size);
+  if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
+    {
+      prot_size = ALIGN_UP (prot_size, GRUB_TXT_PMR_ALIGN);
+
+      if (prot_size > GRUB_TXT_MLE_MAX_SIZE)
+	{
+	  err = GRUB_ERR_OUT_OF_RANGE;
+	  goto fail;
+	}
```
DemiMarie: Should these be reordered to make sure `ALIGN_UP` cannot overflow?

grub-core/loader/i386/linux.c
```diff
@@ -398,6 +500,60 @@ grub_linux_boot_mmap_fill (grub_uint64_t addr, grub_uint64_t size,
...
+    }
+
+  if (boot_params->ramdisk_image)
+      /* TODO the initrd image and size can have hi bits but for now assume always < 4G */
+      grub_slaunch_add_slrt_policy_entry (17,
+                                          GRUB_SLR_ET_RAMDISK,
+                                          /*flags=*/0,
+                                          boot_params->ramdisk_image,
+                                          boot_params->ramdisk_size,
+                                          "Measured Kernel initrd");
```
DemiMarie: Should this be checked?

grub-core/loader/i386/linux.c
```diff
@@ -543,14 +701,17 @@ grub_linux_boot (void)
...
+
   if (! ctx.real_mode_target)
     return grub_error (GRUB_ERR_OUT_OF_MEMORY, "cannot allocate real mode pages");
 
   {
     grub_relocator_chunk_t ch;
     grub_size_t sz;
 
-    if (grub_add (ctx.real_size, efi_mmap_size, &sz))
+    if (grub_add (ctx.real_size, efi_mmap_size + ap_wake_block_size, &sz))
```
DemiMarie: Use `grub_add()` here too?

grub-core/loader/i386/linux.c
```diff
@@ -793,6 +1005,79 @@ grub_cmd_linux (grub_command_t cmd __attribute__ ((unused)),
       goto fail;
     }
 
+  /* Read the kernel_info struct. */
+  if (grub_le_to_cpu16 (lh.version) >= 0x020f)
+    {
+      if (grub_file_seek (file, grub_le_to_cpu32 (lh.kernel_info_offset) +
+				real_size + GRUB_DISK_SECTOR_SIZE) == ((grub_off_t) -1))
```
DemiMarie: `grub_add()`?

grub-core/loader/i386/linux.c
```diff
@@ -793,6 +1005,79 @@ grub_cmd_linux (grub_command_t cmd __attribute__ ((unused)),
...
+	{
+	  grub_error (GRUB_ERR_BAD_OS, N_("incorrect kernel_info header"));
+	  goto fail;
+	}
+
+      if (linux_info->size_total < KERNEL_INFO_MIN_SIZE_TOTAL)
+	{
+	  grub_error (GRUB_ERR_BAD_OS, N_("incorrect kernel_info size"));
+	  goto fail;
+	}
```
DemiMarie, line 1047: Should there be a maximum size too?  The obvious limit is `SSIZE_MAX`, but maybe the limit should be smaller?

grub-core/loader/i386/linux.c
```diff
@@ -793,6 +1005,79 @@ grub_cmd_linux (grub_command_t cmd __attribute__ ((unused)),
...
+	{
+	  if (!grub_errno)
+	    grub_error (GRUB_ERR_BAD_OS, N_("premature end of file %s"), argv[0]);
+	  goto fail;
+	}
+
+      if (grub_slaunch_platform_type () != SLP_NONE)
+	{
+	  if (OFFSET_OF (mle_header_offset, linux_info) >=
+	      grub_le_to_cpu32 (linux_info->size))
```
DemiMarie, line 1070: This should also check that the size of the MLE header is sufficiently large and that the MLE header is valid.

grub-core/loader/i386/slaunch.c
```diff
@@ -0,0 +1,299 @@
...
+                                    grub_uint64_t entity,
+                                    grub_uint64_t size,
+                                    const char *evt_info)
+{
+  struct grub_slr_policy_entry *entry =
+    (void *)((grub_uint8_t *)slr_policy_staging +
+             sizeof(struct grub_slr_entry_policy) +
+             slr_policy_staging->nr_entries*sizeof(*entry));
+
+  if (slr_policy_staging->hdr.size + sizeof(*entry) > sizeof(slr_policy_buf))
```
DemiMarie:
```suggestion
  if (slr_policy_staging->hdr.size > sizeof(slr_policy_buf) - sizeof(*entry))
```
Avoids overflow

grub-core/loader/i386/slaunch.c
```diff
@@ -0,0 +1,299 @@
...
+
+static grub_err_t
+grub_cmd_slaunch_module (grub_command_t cmd __attribute__ ((unused)),
+			 int argc, char *argv[])
+{
+  grub_file_t file;
+  grub_ssize_t size;
+
+  if (!argc)
+    return grub_error (GRUB_ERR_BAD_ARGUMENT, N_("filename expected"));
```
DemiMarie: Should this check for too many arguments?

grub-core/loader/i386/slaunch.c
```diff
@@ -0,0 +1,299 @@
...
+static grub_command_t cmd_slaunch, cmd_slaunch_module, cmd_slaunch_state;
+
+GRUB_MOD_INIT (slaunch)
+{
+  cmd_slaunch = grub_register_command ("slaunch", grub_cmd_slaunch,
+				       NULL, N_("Enable secure launcher"));
+  cmd_slaunch_module = grub_register_command ("slaunch_module", grub_cmd_slaunch_module,
+					      NULL, N_("Load secure launcher module from file"));
+  cmd_slaunch_state = grub_register_command ("slaunch_state", grub_cmd_slaunch_state,
+					     NULL, N_("Display secure launcher state"));
```
DemiMarie, line 288: Are these allowed to fail?

grub-core/loader/i386/txt/acmod.c
```diff
@@ -0,0 +1,610 @@
...
+		     proc_id_list->count ) == NULL )
+    {
+      grub_error (GRUB_ERR_OUT_OF_RANGE, N_("ACM processor ID entries out of bounds"));
+      return NULL;
+    }
+
+  return proc_id_list;
+}
+
+static int
```
DemiMarie, line 246:
```suggestion
static bool
```

include/grub/i386/slaunch.h
```diff
@@ -0,0 +1,85 @@
...
+  grub_uint32_t mle_ptab_size;
+  void *mle_ptab_mem;
+  grub_uint32_t mle_header_offset;
+  grub_uint32_t ap_wake_block;
+  grub_uint32_t ap_wake_block_size;
+  grub_uint32_t dce_base;
+  grub_uint32_t dce_size;
+  grub_uint64_t tpm_evt_log_base;
+  grub_uint32_t tpm_evt_log_size;
+};
```
DemiMarie, line 71: This should either be `__attribute__((packed))` or reordered to eliminate padding.

include/grub/i386/msr.h
```diff
@@ -0,0 +1,137 @@
...
+#define GRUB_MSR_X86_MTRR_DEF_TYPE	0x000002ff
+#define GRUB_MSR_X86_DEF_TYPE_MASK	0xff
+#define GRUB_MSR_X86_MTRR_ENABLE_FIXED	(1<<10)
+#define GRUB_MSR_X86_MTRR_ENABLE	(1<<11)
+
+#define GRUB_MSR_X86_MC0_STATUS		0x00000401
+
+#define GRUB_MSR_X86_EFER		0xc0000080 /* Extended features */
+#define GRUB_MSR_EFER_LME		(1<<8)     /* Enable Long Mode/IA-32e */
+#define GRUB_MSR_EFER_LMA		(1<<10)    /* Long Mode/IA-32e Active */
```
DemiMarie, line 67: These are not documented by Intel.  Are they AMD-specific?

grub-core/loader/i386/txt/acmod.c
```diff
@@ -0,0 +1,610 @@
...
+ * NULL is returned instead.
+ */
+static struct grub_txt_acm_info_table*
+get_acmod_info_table (struct grub_txt_acm_header* hdr)
+{
+  grub_uint32_t user_area_off, info_table_size;
+  struct grub_txt_acm_info_table *ptr = NULL;
+  /* Minimum size required to read full size of table */
+  info_table_size = __builtin_offsetof (struct grub_txt_acm_info_table, length)
+		    + sizeof(ptr->length);
```
DemiMarie, line 133:
```suggestion
  info_table_size = offsetof (struct grub_txt_acm_info_table, length)
		    + sizeof(ptr->length);
```

grub-core/loader/i386/txt/acmod.c
```diff
@@ -0,0 +1,610 @@
...
+      return NULL;
+    }
+
+  /* Overflows? */
+  if ( grub_add (info_table->processor_id_list, sizeof(*proc_id_list),
+                 &id_entries_off))
+    {
+      grub_error (GRUB_ERR_OUT_OF_RANGE, N_("ACM processor ID entries out of bounds"));
+      return NULL;
+    }
```
DemiMarie: `fits_in_acm()` checked this already
```suggestion
  id_entries_off = info_table->processor_id_list + sizeof(*proc_id_list);
```

grub-core/loader/i386/txt/acmod.c
```diff
@@ -0,0 +1,610 @@
...
+
+  /* Overflows? */
+  if ( grub_add (info_table->chipset_id_list,
+		 sizeof(struct grub_txt_acm_chipset_id_list), &id_entries_off))
+    {
+      grub_error (GRUB_ERR_OUT_OF_RANGE, N_("ACM chipset ID entries out of bounds"));
+      return NULL;
+    }
```
DemiMarie: `fits_in_acm()` checked this already.

```suggestion
  id_entries_off = info_table->chipset_id_list + sizeof(struct grub_txt_acm_chipset_id_list);
```

grub-core/loader/i386/txt/acmod.c
```diff
@@ -0,0 +1,610 @@
...
+       * supported by this code.
+       *
+       * TXT spec says that 9 will be the final version and further changes will
+       * be reflected elsewhere, but check for higher values too in case they
+       * change their mind.
+       */
+      if ( info_table->version >= 9 )
+	grub_error (GRUB_ERR_NOT_IMPLEMENTED_YET,
+		    N_("chipset id mismatch, flexible ACM info list may contain"
+		       " matching entry but it isn't yet supported by code"));
```
DemiMarie, line 478: should this be fixed?

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+#define GRUB_SLR_ET_SLRT			0x0001
+#define GRUB_SLR_ET_BOOT_PARAMS			0x0002
+#define GRUB_SLR_ET_SETUP_DATA			0x0003
+#define GRUB_SLR_ET_CMDLINE			0x0004
+#define GRUB_SLR_ET_UEFI_MEMMAP			0x0005
+#define GRUB_SLR_ET_RAMDISK			0x0006
+#define GRUB_SLR_ET_MULTIBOOT2_INFO		0x0007
+#define GRUB_SLR_ET_MULTIBOOT2_MODULE		0x0008
+#define GRUB_SLR_ET_TXT_OS2MLE			0x0010
+#define GRUB_SLR_ET_UNUSED			0xffff
```
DemiMarie, line 80: Should there be additional entries for microcode and dom0?

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+  struct grub_slr_txt_mtrr_state saved_bsp_mtrrs;
+} GRUB_PACKED;
+
+/*
+ * AMD SKINIT Info table
+ */
+struct grub_slr_entry_amd_info
+{
+  struct grub_slr_entry_hdr hdr;
+} GRUB_PACKED;
```
DemiMarie:
```suggestion
struct grub_slr_entry_amd_info
{
  struct grub_slr_entry_hdr hdr;
  char reserved[]; /* Reserved for future use */
} GRUB_PACKED;
```

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+
+/*
+ * ARM DRTM Info table
+ */
+struct grub_slr_entry_arm_info
+{
+  struct grub_slr_entry_hdr hdr;
```
DemiMarie, line 204:
```suggestion
  struct grub_slr_entry_hdr hdr;
  char reserved[]; /* Reserved for future use */
```

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+} GRUB_PACKED;
+
+struct grub_slr_uefi_cfg_entry
+{
+  grub_uint16_t pcr;
+  grub_uint16_t reserved;
+  grub_uint64_t cfg; /* address or value */
+  grub_uint32_t size;
+  char evt_info[GRUB_TPM_EVENT_INFO_LENGTH];
+} GRUB_PACKED;
```
DemiMarie, line 215: If this struct is variable length, it should have a flexable array member (even if merely `char reserved[];`) to indicate that.  Otherwise, it should be included explicitly in `struct grub_slr_entry_uefi_config`.

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+
+static inline struct grub_slr_entry_hdr *
+grub_slr_next_entry (struct grub_slr_table *table,
+                     struct grub_slr_entry_hdr *curr)
+{
+  struct grub_slr_entry_hdr *next = (struct grub_slr_entry_hdr *)
+                                    ((grub_uint8_t *) curr + curr->size);
+
+  if ((void *)next >= grub_slr_end_of_entries (table))
+    return NULL;
```
DemiMarie: Out-of-bounds pointer arithmetic is undefined behavior, so the bounds check needs to come _before_ the pointer arithmetic.  With `-fno-strict-overflow` it is not undefined, but it is still a bad idea and risks wrapping.  The only exception is if this will only ever run on trusted input.

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+  return next;
+}
+
+static inline struct grub_slr_entry_hdr *
+grub_slr_next_entry_by_tag (struct grub_slr_table *table,
+                            struct grub_slr_entry_hdr *entry,
+                            grub_uint16_t tag)
+{
+  if (!entry) /* Start from the beginning */
+    entry = (struct grub_slr_entry_hdr *)((grub_uint8_t *) table + sizeof(*table));
```
DemiMarie, line 273: This should check that the first entry is valid (fits in the table), again unless input is known trusted here.

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+{
+  struct grub_slr_entry_hdr *next = (struct grub_slr_entry_hdr *)
+                                    ((grub_uint8_t *) curr + curr->size);
+
+  if ((void *)next >= grub_slr_end_of_entries (table))
+    return NULL;
+  if (next->tag == GRUB_SLR_ENTRY_END)
+    return NULL;
+
+  return next;
```
DemiMarie, line 264: This should check that `next` is valid, subject to the same caveat about input being trusted.

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+  return NULL;
+}
+
+static inline int
+grub_slr_add_entry (struct grub_slr_table *table,
+                    struct grub_slr_entry_hdr *entry)
+{
+  struct grub_slr_entry_hdr *end;
+
+  if ((table->size + entry->size) > table->max_size)
```
DemiMarie: `grub_add()`?

include/grub/slr_table.h
```diff
@@ -0,0 +1,298 @@
...
+  end = (struct grub_slr_entry_hdr *)((grub_uint8_t *) table + table->size - sizeof(*end));
+  end->tag = GRUB_SLR_ENTRY_END;
+  end->size = sizeof(*end);
+
+  return 0;
+}
+
+static inline void
+grub_slr_init_table(struct grub_slr_table *slrt, grub_uint16_t architecture,
+                    grub_uint32_t max_size)
```
DemiMarie, line 312: This should check that `max_size` is large enough, and `grub_fatal()` if it is not.

include/grub/i386/txt.h
```diff
@@ -0,0 +1,748 @@
...
+                                           grub_txt_bios_data_size (heap) +
+                                           grub_txt_os_mle_data_size (heap) +
+                                           grub_txt_os_sinit_data_size (heap) +
+                                           sizeof (grub_uint64_t));
+}
+
+/* Intel 64 and IA-32 Architectures Software Developer’s Manual */
+/* Volume 2 (2A, 2B, 2C & 2D): Instruction Set Reference, A-Z */
+
+/* CHAPTER 6 SAFER MODE EXTENSIONS REFERENCE */
```
DemiMarie:
```suggestion
/* CHAPTER 7 SAFER MODE EXTENSIONS REFERENCE */
```

include/grub/i386/txt.h
```diff
@@ -0,0 +1,748 @@
...
+
+#define GRUB_SMX_LEAF_CAPABILITIES	0
+#define GRUB_SMX_LEAF_UNDEFINED		1
+#define GRUB_SMX_LEAF_ENTERACCS		2
+#define GRUB_SMX_LEAF_EXITAC 		3
+#define GRUB_SMX_LEAF_SENTER		4
+#define GRUB_SMX_LEAF_SEXIT		5
+#define GRUB_SMX_LEAF_PARAMETERS	6
+#define GRUB_SMX_LEAF_SMCTRL		7
+#define GRUB_SMX_LEAF_WAKEUP		8
```
DemiMarie, line 576: This is Table 7-2.

include/grub/i386/txt.h
```diff
@@ -0,0 +1,748 @@
...
+  asm volatile ("getsec" :
+                        : "a" (GRUB_SMX_LEAF_SENTER),
+                          "b" (acm_phys_addr), "c" (acm_size), "d"(0));
+}
+
+static inline void
+grub_txt_getsec_sexit (void)
+{
+  asm volatile ("getsec" : : "a" (GRUB_SMX_LEAF_SEXIT));
+}
```
DemiMarie:
```suggestion
```

This is for the measured environment to use, not GRUB.

include/grub/i386/txt.h
```diff
@@ -0,0 +1,748 @@
...
+                        : "a" (GRUB_SMX_LEAF_ENTERACCS),
+                          "b" (acm_phys_addr), "c" (acm_size));
+}
+
+static inline void
+grub_txt_getsec_exitac (grub_uint32_t near_jump)
+{
+  asm volatile ("getsec" :
+                        : "a" (GRUB_SMX_LEAF_EXITAC), "b" (near_jump));
+}
```
DemiMarie:
```suggestion
```
This function is unusable code: outside of the ACM (which is signed by Intel and therefore will never be provided by GRUB) it will always raise `#GP(0)`.

include/grub/i386/txt.h
```diff
@@ -0,0 +1,748 @@
...
+#define GRUB_SMX_GET_TXT_EXT_FEATURES(v) (v & (__typeof__(v))(GRUB_SMX_PROCESSOR_BASE_SCRTM|GRUB_SMX_MACHINE_CHECK_HANDLING))
+
+#define GRUB_SMX_DEFAULT_VERSION	0x0
+#define GRUB_SMX_DEFAULT_VERSION_MASK	0xffffffff
+#define GRUB_SMX_DEFAULT_MAX_ACM_SIZE	0x8000 /* 32K */
+#define GRUB_SMX_DEFAULT_ACM_MEMORY_TYPE GRUB_SMX_ACM_MEMORY_TYPE_UC
+#define GRUB_SMX_DEFAULT_SENTER_CONTROLS 0x0
+
+#define GRUB_TXT_PMR_ALIGN_SHIFT	21
+#define GRUB_TXT_PMR_ALIGN		(1 << GRUB_TXT_PMR_ALIGN_SHIFT)
```
DemiMarie, line 650: Source for this?

include/grub/i386/mmio.h
```diff
@@ -0,0 +1,72 @@
...
+grub_write32 (grub_uint32_t val, grub_addr_t addr)
+{
+  (*(volatile grub_uint32_t *) (addr)) = val;
+}
+
+static inline void
+grub_write64 (grub_uint64_t val, grub_addr_t addr)
+{
+  (*(volatile grub_uint64_t *) (addr)) = val;
+}
```
DemiMarie, line 70: I recommend using inline assembly here.  These functions are x86-specific, so there is no loss in portability, and there is a _gain_ from specifying the exact instructions that will be used for MMIO.  This is especially important in virtualized environments, as some VMMs do not support emulation of certain instructions.

This aligns with my Rust RFC for `core::arch::{load, store}`, which explicitly describes the assembler instructions to be used.

include/grub/i386/linux.h
```diff
@@ -148,6 +148,17 @@ struct linux_i386_kernel_header
...
+} GRUB_PACKED;
+
+struct linux_kernel_info
+{
+  grub_uint32_t header;
+  grub_uint32_t size;
+  grub_uint32_t size_total;
+  grub_uint32_t setup_type_max;
+  grub_uint32_t mle_header_offset;
+  grub_uint8_t  var_len_data[0];
```
DemiMarie:
```suggestion
  grub_uint8_t  var_len_data[];
```

include/grub/i386/linux.h
```diff
@@ -148,6 +148,17 @@ struct linux_i386_kernel_header
...
   grub_uint32_t init_size;
   grub_uint32_t handover_offset;
+  grub_uint32_t kernel_info_offset;
+} GRUB_PACKED;
+
+struct linux_kernel_info
+{
+  grub_uint32_t header;
+  grub_uint32_t size;
+  grub_uint32_t size_total;
```
DemiMarie: What are the units of these fields?

grub-core/mmap/mmap.c
```diff
@@ -343,6 +344,84 @@ grub_mmap_unregister (int handle)
...
+   the passed limit. If no RAM exists above the limit, ~0 is returned. */
+grub_uint64_t
+grub_mmap_get_lowest (grub_uint64_t limit)
+{
+  addr_limit_t al = {~0, limit};
+
+  grub_mmap_iterate (lowest_hook, &al);
+
+  return al.addr;
+}
```
DemiMarie, line 386: Should this function also take a size argument?  I am rather surprised that any caller needs only 1 byte of RAM at this location.

grub-core/mmap/mmap.c
```diff
@@ -343,6 +344,84 @@ grub_mmap_unregister (int handle)
...
+   returned. */
+grub_uint64_t
+grub_mmap_get_highest (grub_uint64_t limit)
+{
+  addr_limit_t al = {0, limit};
+
+  grub_mmap_iterate (highest_hook, &al);
+
+  return al.addr;
+}
```
DemiMarie, line 423: This function should also take a size argument.

grub-core/loader/i386/txt/txt.c
```diff
@@ -0,0 +1,1113 @@
...
+    {
+      grub_dprintf ("slaunch", "Firmware didn't lock FEATURE_CONTROL MSR,"
+		    "locking it now\n");
+      /* Not setting SENTER_FUNCTIONS and SENTER_ENABLE because they were tested
+       * in grub_txt_verify_platform() */
+      feat_ctrl |= GRUB_MSR_X86_ENABLE_VMX_OUT_SMX |
+                   GRUB_MSR_X86_ENABLE_VMX_IN_SMX |
+                   GRUB_MSR_X86_FEATURE_CTRL_LOCK;
+      grub_wrmsr (GRUB_MSR_X86_FEATURE_CONTROL, feat_ctrl);
+    }
```
DemiMarie, line 97: Is the lock to make sure that no other code can disable these?

grub-core/loader/i386/txt/txt.c
```diff
@@ -0,0 +1,1113 @@
...
+    {
+      grub_error (GRUB_ERR_BAD_DEVICE, N_("GETSEC[SENTER] is not available"));
+      goto fail;
+    }
+
+  if (!(caps & GRUB_SMX_CAPABILITY_PARAMETERS))
+    {
+      grub_error (GRUB_ERR_BAD_DEVICE, N_("GETSEC[PARAMETERS] is not available"));
+      goto fail;
+    }
```
DemiMarie, line 120: Why is this an error?  Is it because it is not expected to ever happen in practice?

grub-core/loader/i386/txt/txt.c
```diff
@@ -0,0 +1,1113 @@
...
+        params->senter_controls = GRUB_SMX_GET_SENTER_CONTROLS (eax);
+        break;
+
+      case GRUB_SMX_PARAMETER_TXT_EXTENSIONS:
+        params->txt_feature_ext_flags = GRUB_SMX_GET_TXT_EXT_FEATURES (eax);
+        break;
+
+      default:
+	grub_error (GRUB_ERR_BAD_ARGUMENT, N_("Unknown SMX parameter"));
+	param_type = GRUB_SMX_PARAMETER_NULL;
```
DemiMarie: Why is this an error?  Future CPUs might expose more parameters.

grub-core/loader/i386/txt/txt.c
```diff
@@ -0,0 +1,1113 @@
...
+      case GRUB_SMX_PARAMETER_ACM_MEMORY_TYPES:
+        params->acm_memory_types = GRUB_SMX_GET_ACM_MEMORY_TYPES (eax);
+        break;
+
+      case GRUB_SMX_PARAMETER_SENTER_CONTROLS:
+        params->senter_controls = GRUB_SMX_GET_SENTER_CONTROLS (eax);
+        break;
+
+      case GRUB_SMX_PARAMETER_TXT_EXTENSIONS:
+        params->txt_feature_ext_flags = GRUB_SMX_GET_TXT_EXT_FEATURES (eax);
```
DemiMarie, line 172: Does this require a loop to get all the extensions?

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,269 @@
...
+verify_acm_elt (struct grub_txt_heap_ext_data_element *elt)
+{
+  grub_uint64_t acm_addrs_size;
+  grub_uint32_t i;
+
+  if ( elt->size < GRUB_TXT_HEAP_ELEMENT_HEADER_SIZE +
+                   sizeof(struct grub_txt_heap_acm_element) ||
+       grub_mul (elt->acm.num_acms, sizeof(grub_uint64_t), &acm_addrs_size) ||
+       elt->size != GRUB_TXT_HEAP_ELEMENT_HEADER_SIZE + sizeof(elt->acm) +
+                    acm_addrs_size )
```
DemiMarie:
```suggestion
       elt->size - (GRUB_TXT_HEAP_ELEMENT_HEADER_SIZE + sizeof(elt->acm)) !=
                    acm_addrs_size )
```

avoids integer overflow

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,269 @@
...
+  grub_err_t err;
+
+  if ( elts_size < GRUB_TXT_HEAP_ELEMENT_HEADER_SIZE )
+     return grub_error (GRUB_ERR_OUT_OF_RANGE,
+                 N_("TXT heap ext data elements too small"));
+
+  for ( ; ; )
+    {
+      if ( elts_size < GRUB_TXT_HEAP_ELEMENT_HEADER_SIZE ||
+           elts_size < elt->size || elt->size == 0 )
```
DemiMarie:
```suggestion
           elts_size < elt->size || elt->size < GRUB_TXT_HEAP_ELEMENT_HEADER_SIZE )
```

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,269 @@
...
+    return grub_error (GRUB_ERR_BAD_DEVICE, N_("GETSEC[SENTER] is not enabled"));
+
+  /*
+   * TODO
+   * TXT Specification
+   * 4.5 SGX Requirement for TXT Platform
+   * Secure Launch currently does not support interop with SGX since it does
+   * not have TPM support to write the SE NVRAM index.
+   * Eventually need the verify_IA32_se_svn_status routine to be called here.
+   */
```
DemiMarie, line 220: That’s fine for Qubes OS, as we don’t support SGX.

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,269 @@
...
+
+  errorcode = grub_txt_reg_pub_readl (GRUB_TXT_ERRORCODE);
```
DemiMarie: This function should be renamed `grub_txt_reg_pub_read32`.  Similarly for the other register read/write functions.

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,269 @@
...
+			N_("TXT_ERRORCODE reports failure: 0x%08" PRIxGRUB_UINT32_T),
+			errorcode);
+
+  if (grub_txt_reg_pub_readb (GRUB_TXT_ESTS) & GRUB_TXT_ESTS_TXT_RESET)
+     return grub_error (GRUB_ERR_BAD_DEVICE,
+			N_("TXT_RESET.STS is set and GETSEC[SENTER] is disabled"));
+
+  /*
+   * Verify that the BIOS information in the TXT heap that was setup by the
+   * BIOS ACM is sane.
```
DemiMarie:
```suggestion
   * BIOS ACM is reasonable.
```

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,269 @@
...
+  if ( bios_data->num_logical_procs > TXT_MAX_CPUS )
+     return grub_error (GRUB_ERR_OUT_OF_RANGE,
+                 N_("BIOS reports too many CPUs for secure launch (%d)"),
+                 bios_data->num_logical_procs);
+
+  if ( bios_data->version >= 4 && bios_size > sizeof(*bios_data) + sizeof(bios_size) )
+    {
+      elts = (struct grub_txt_heap_ext_data_element *) ((grub_uint8_t *)bios_data + sizeof(*bios_data));
+      err = verify_ext_data_elts(elts, bios_size - sizeof(*bios_data));
+    }
```
DemiMarie: What if the version is 3, or the size is too small?  At least the latter should be an error.

grub-core/loader/multiboot_elfxx.c
```diff
@@ -180,6 +247,27 @@ CONCAT(grub_multiboot_load_elf, XX) (mbi_load_data_t *mld)
...
+
+      /* TODO: decide on universal way of conveying location of MLE header */
+      for (mle_hdr_offset = 0;  mle_hdr_offset < 0x1000; mle_hdr_offset += 16)
+	{
+	  if ( !grub_memcmp ((void *)((grub_addr_t) source + mle_hdr_offset), GRUB_TXT_MLE_UUID, 16) )
+	    {
+	      slparams->mle_header_offset = mle_hdr_offset;
+	      break;
+	    }
+	}
```
DemiMarie, line 262: A solution that did not involve scanning would be nice.  Also, this needs to avoid reading out of bounds.

grub-core/loader/multiboot_mbi2.c
```diff
@@ -1126,3 +1168,62 @@ grub_multiboot2_set_bootdev (void)
...
+
+static void
+add_multiboot2_slrt_policy_entries (void)
+{
+  unsigned i;
+  struct module *cur;
+
+  for (i = 0, cur = modules; i < modcnt; i++, cur = cur->next)
+    {
+      grub_slaunch_add_slrt_policy_entry (17,
```
DemiMarie, line 1180: Where does 17 come from?

grub-core/loader/i386/txt/txt.c
```diff
@@ -0,0 +1,1113 @@
...
+void
+grub_txt_add_slrt_policy_entries (void)
+{
+  struct grub_txt_os_mle_data *os_mle_data;
+  grub_uint8_t *txt_heap;
+
+  txt_heap = grub_txt_get_heap ();
+  os_mle_data = grub_txt_os_mle_data_start (txt_heap);
+
+  grub_slaunch_add_slrt_policy_entry (18,
```
DemiMarie: This magic number needs to be documented somewhere.

grub-core/loader/i386/slaunch.c
```diff
@@ -0,0 +1,299 @@
...
+  slr_policy_staging->hdr.tag = GRUB_SLR_ENTRY_DRTM_POLICY;
+  slr_policy_staging->hdr.size = sizeof(struct grub_slr_entry_policy);
+  slr_policy_staging->revision = GRUB_SLR_TABLE_REVISION;
+  slr_policy_staging->nr_entries = 0;
+}
+
+void grub_slaunch_add_slrt_policy_entries (void)
+{
+  /* The SLR table should be measured too, at least parts of it. */
+  grub_slaunch_add_slrt_policy_entry (18,
```
DemiMarie: Magic number needs to be documented.

DemiMarie (changes requested):
Please rename all `read` and `write` functions to include the number of bits they access, so `read64` instead of `readq` (and similar for other functions).  Also please document the magic numbers used in the SLRT calls

docs/grub.texi
krystian-hebel: Yes, from the description above:

> (...) GRUB will log each command executed and each file loaded into
the TPM event log and extend the PCR values in the TPM correspondingly.

This isn't in any way modified by Secure Launch implemented by this PR. All the measuring introduced by SL takes place after GRUB.

grub-core/lib/i386/relocator32.S
krystian-hebel: It won't, there is a handcrafted `jmpl ptr16:32` in between. There is no way of returning here from the code jumped to.

Short introduction on how relocator works, only parts important for this discussion:

GRUB is loaded somewhere in the memory space. Whatever loads GRUB doesn't know what will be loaded next, so it can't decide on location based on that. GRUB then may be ordered to load a non-relocatable kernel/module/whatever to a memory overlapping with GRUB itself, either its code or data. In that case it notes this fact, but keeps the loaded thing wherever it found free space. This is done by any of the `grub_relocator_alloc_chunk*()` functions, the chunk has "virtual" current and "physical" target addresses pair (in fact, both addresses are physical, hence the quote marks). Usually there is about a dozen or two of chunks of memory to keep track of, but it depends on what is actually loaded.

At the end of the day, GRUB takes all the chunks, sorts them, creates a list of chunks that need to be moved. This is basically a list of arguments to `memmove()`, with one critical caveat: `memmove()` itself may have to be overwritten in the process. So, instead of using calls to possibly non-existing function, GRUB allocates another chunk, big enough to hold the code that does the relocation. The code it then created at runtime from a template ("compiled") to something that looks like a set of inlined function calls. All of the addresses are hardcoded into the instructions - any data outside of that chunk may already be overwritten at this point.

Jumping into this set of "memmoves" is a point of no return - all of the GRUB code (except from that single chunk) and data is treated as gone. For this reason, after the "memmoves" there is another set of instructions that loads the registers and jumps to the kernel. This is what happens in this file, or its 16/64-bit counterparts for different boot protocols. For the same reasons as before, this must be true position independent code, with no references to data. All the values are thus loaded directly from the instructions, mostly as part of `mov imm32, %reg`.

Neither register values nor entry point of kernel are known at the compile time, they must be filled by the code. One option would be to create a label before the instruction and manually add 1 byte (or 2 for _some_ of the 64b instructions) for opcode to the address, but that would be atrocious. Instead, a lot of instructions in this file is handcrafted (`.byte`, `.long`, occasional `.word`). This is done so that the expected immediate value for given opcodes can be labeled and referenced from C code as external variables.

grub-core/loader/i386/linux.c
krystian-hebel: Added cast to `grub_off_t` instead, which is defined as `grub_uint64_t`. `real_size + GRUB_DISK_SECTOR_SIZE` should never overflow, `real_size` comes from uint8_t limited to `GRUB_LINUX_MAX_SETUP_SECTS` = 64 and then shifted 9 bits to the left, second argument is 512.

grub-core/loader/i386/linux.c
krystian-hebel: [Kernel documentation](https://www.kernel.org/doc/html/latest/arch/x86/boot.html) doesn't specify any limit, but I have to ask: why signed?

krystian-hebel: Also, this is currently [being reviewed on Linux side](https://groups.google.com/g/trenchboot-devel/c/7sYkzOWuDro/m/0LBmdRCQAAAJ) and as far as I can see, it already was changed. Version 8 of Linux patches will be ready soon (this week or the next one). I'm considering dropping it from this PR since Xen doesn't use it anyway, but then it may get even harder to keep everything aligned.

grub-core/loader/i386/slaunch.c
krystian-hebel: Only in the unlikely event that there is no memory to allocate new element on the list of commands. I added checks in `GRUB_MOD_FINI` below to avoid dereferencing null pointer.

grub-core/loader/i386/txt/acmod.c
krystian-hebel: `offsetof` is a macro defined in `stddef.h`, which isn't used in freestanding code like this one.

grub-core/loader/i386/txt/acmod.c
krystian-hebel: No such type in freestanding code before C23.

grub-core/loader/i386/txt/acmod.c
krystian-hebel: We have no way of testing it.

grub-core/loader/i386/txt/txt.c
krystian-hebel: Yes. In addition, SINIT ACM is supposed to throw an error if this register isn't locked, although I haven't got the chance to check if that's true.

grub-core/loader/i386/txt/txt.c
krystian-hebel: More because it is used by `grub_txt_smx_parameters()`.

grub-core/loader/i386/txt/txt.c
krystian-hebel: No, according to the description in SDM there are only 2 meaningful bits returned in %eax.

grub-core/loader/i386/txt/txt.c
krystian-hebel: SDM on `GETSEC[PARAMETERS]`:

> Table 7-7 defines the parameter types supported in current **and future** implementations.

By my understanding that list is final.

grub-core/loader/i386/txt/verify.c
krystian-hebel: Version 3 has no extended data elements, so there is nothing to verify. If the size is too small, there are no extended data elements and there is nothing to verify.

Size is checked earlier, I updated it to check for proper size instead of just uint64_t.

grub-core/mmap/mmap.c
krystian-hebel: These functions aren't used to allocate or map memory. They were added to **find** a range in memory map (mmap) that has the highest/lowest RAM (not MMIO) address. This is used to set protection against DMA. The returned addresses aren't accessed.

include/grub/i386/msr.h
krystian-hebel: This is the way CPU goes into 64-bit mode, it is documented.

image

include/grub/i386/slaunch.h
krystian-hebel: Why is padding a problem here?

include/grub/i386/txt.h
krystian-hebel: Removed also next function, so nobody gets tempted to invoke SENTER before relocator does its job.

include/grub/slr_table.h
krystian-hebel: Those are passed as `MULTIBOOT2_MODULE`. GRUB doesn't parse the contents of modules, other than optional decompression.

include/grub/slr_table.h
krystian-hebel: Added check against size of the table to `grub_slr_init_table()`, it includes one entry so it should be enough to assume that at least first entry is valid.

grub-core/loader/i386/linux.c
```diff
@@ -181,14 +181,15 @@ allocate_pages (grub_size_t prot_size, grub_size_t *align,
 
   if (grub_slaunch_platform_type () == SLP_INTEL_TXT)
     {
-      prot_size = ALIGN_UP (prot_size, GRUB_TXT_PMR_ALIGN);
-
       if (prot_size > GRUB_TXT_MLE_MAX_SIZE)
```
DemiMarie, line 184: Should there be a compile-time check that `GRUB_TXT_MLE_MAX_SIZE` is unchanged when rounded up by `GRUB_TXT_PMR_ALIGN`?

grub-core/loader/i386/txt/acmod.c
DemiMarie: I thought `stddef.h` _is_ available in freestanding code.  On my system (GCC with Fedora 39) it is provided by the compiler, not libc, and it only provides definitions for macros, structs, and typedefs that are defined by the compiler.

grub-core/loader/i386/txt/acmod.c
DemiMarie: `typedef _Bool bool`?  That said, I thought `<stdbool.h>` was available even in freestanding mode.  You don’t get a libc in freestanding mode, but `<stdbool.h>` is the compiler’s job to provide, not libc’s.

grub-core/loader/i386/txt/acmod.c
DemiMarie: Are there devices with versions greater than 8?

grub-core/loader/i386/txt/verify.c
```diff
@@ -218,32 +219,35 @@ grub_txt_verify_platform (void)
...
   if ( txt_heap == NULL || heap_base == 0 || heap_size == 0 )
      return grub_error (GRUB_ERR_BAD_DEVICE,
                  N_("TXT heap is not configured correctly"));
 
   bios_size = grub_txt_bios_data_size (txt_heap);
-  if ( bios_size < sizeof (grub_uint64_t) || bios_size > heap_size )
+  /* We support versions >= 4, but bios_data->mle_flags is in versions >= 5. */
+  if ( bios_size < sizeof (grub_uint64_t) + sizeof (*bios_data) -
+                   sizeof (bios_data->mle_flags) ||
+       bios_size > heap_size )
```
DemiMarie: Would this be a good use for `__builtin_offsetof()`?

include/grub/i386/linux.h
```diff
@@ -154,11 +154,11 @@ struct linux_i386_kernel_header
 struct linux_kernel_info
 {
   grub_uint32_t header;
-  grub_uint32_t size;
-  grub_uint32_t size_total;
+  grub_uint32_t size;		/* In bytes, excluding var_len_data[] */
+  grub_uint32_t size_total;	/* In bytes, including var_len_data[] */
```
DemiMarie, line 158: Are these redundant with each other, or could the offset of `var_len_data` change in the future?

include/grub/slr_table.h
```diff
@@ -285,6 +313,9 @@ grub_slr_init_table(struct grub_slr_table *slrt, grub_uint16_t architecture,
 {
   struct grub_slr_entry_hdr *end;
 
+  if (max_size < sizeof(*slrt) + sizeof(*end))
+    grub_fatal ("Requested SLRT max size (%d B) is too small\n", max_size);
```
DemiMarie:
```suggestion
    grub_fatal ("Requested SLRT max size (%u B) is too small\n", max_size);
```

or use the appropriate `GRUB_PRIu32` macro.

include/grub/slr_table.h
```diff
@@ -218,21 +214,50 @@ struct grub_slr_uefi_cfg_entry
...
                      struct grub_slr_entry_hdr *curr)
 {
-  struct grub_slr_entry_hdr *next = (struct grub_slr_entry_hdr *)
-                                    ((grub_uint8_t *) curr + curr->size);
+  grub_addr_t addr;
+  struct grub_slr_entry_hdr *next;
+
+  /* Can read the size field of current entry? */
+  if ( grub_add ((grub_addr_t) curr, sizeof(*curr), &addr) )
+    return NULL;
```
DemiMarie, line 240: I suggest having a `grub_slr_first_entry()` function that validates that the first entry is valid.  Then `grub_slr_next_entry()` can either assume or assert that `curr` is valid.  By mathematical induction, this is enough to validate all entries in the table.

Here, this check is redundant, but you need to check that first `table->size` and then `curr->size` are at least `sizeof(*cur)`.

```suggestion
  /* Can read the size field of current entry? */
  if ( table->size < sizeof(*curr) )
    return NULL;

 /* Is the current entry large enough? */
 if ( curr->size < sizeof(*curr) )
   return NULL;
```

include/grub/slr_table.h
```diff
@@ -218,21 +214,50 @@ struct grub_slr_uefi_cfg_entry
...
+
+  /* Does next element's header fit within the table? */
+  if (addr >= grub_slr_end_of_entries (table))
+    return NULL;
+
+  next = (struct grub_slr_entry_hdr *) (addr - sizeof(*next));
+
+  /* Does next element fit within the table? */
+  if (grub_slr_end_of_entries (table) - (addr - sizeof(*next)) < next->size)
+    return NULL;
```
DemiMarie, line 258:
```suggestion
  /* Does next element's header fit within the table? */
  if (addr >= grub_slr_end_of_entries (table) - sizeof(*next))
    return NULL;

  next = (struct grub_slr_entry_hdr *) addr);

  /* Does next element fit within the table? */
  if (grub_slr_end_of_entries (table) - addr < next->size)
    return NULL;

  /* Does next element include a header? */
  if (next->size < sizeof(*next))
    return NULL;
```

grub-core/loader/i386/txt/acmod.c
krystian-hebel: I find it hard to believe that Intel publicized documentation unless such hardware exists. That said, ACMs in the official package ([630744](https://cdrdv2.intel.com/v1/dl/getContent/630744)) don't even have version 8:

```
$ for f in *.bin; do xxd $f | grep "aa3a c07f a746 db18" -A1 | tail -n1 | cut -b13-14 -z; echo " $f"; done
07 ADL_SINIT_v1_18_16_20230427_REL_NT_O1.PW_signed.bin
05 BDW_SINIT_20190708_1.3.2_PW.bin
06 CFL_SINIT_20221220_PRODUCTION_REL_NT_O1_1.10.1_signed.bin
07 CML_RKL_S_SINIT_v1.13.33_REL_NT_O1.PW_signed.bin
07 CML_S_SINIT_1_13_33_REL_NT_O1.PW_signed.bin
07 CMLSTGP_SINIT_v1_14_46_20220819_REL_NT_O1.PW_signed.bin
07 RKLS_SINIT_v1_14_46_20220819_REL_NT_O1.PW_signed.bin
06 SKL_KBL_AML_SINIT_20211019_PRODUCTION_REL_NT_O1_1.10.0.bin
04 SNB_IVB_SINIT_20190708_PW.bin
07 TGL_SINIT_v1_14_46_20220819_REL_NT_O1.PW_signed.bin
```

grub-core/loader/i386/linux.c
krystian-hebel: These are defined in a way that doesn't make it possible to provide different values without changing the header file. There is no additional security in adding a compile-time assertion, because removing that assertion wouldn't require anything that changing the values requires. In both cases attacker would have to modify the code before compilation.

grub-core/loader/i386/txt/acmod.c
krystian-hebel: You're right, I've mistook it with `stdint.h`, which in theory is available since C99, but compilers still have problems with it.

In any case, GRUB seems to avoid standard headers in core code, except for `stdarg.h`. They are used in OS utilities and libraries, but not in the bootloader itself, so I don't feel comfortable including it just for this.

grub-core/loader/i386/txt/acmod.c
krystian-hebel: Same reasoning as above.

include/grub/i386/linux.h
krystian-hebel: `var_len_data` has variable length, this length is included only in `size_total`. This is described in https://www.kernel.org/doc/html/latest/arch/x86/boot.html#details-of-the-kernel-info-fields, but in my opinion that description is unnecessarily complicated. `size` is and will be (in case other fields are added) offset to `var_len_data` and that's everything to know.

In fact, newest Linux counterpart has MLE header offset moved from `var_len_data` to `kernel_info` proper, due to issues with calculating relative addresses between compilation units. This code will have to be updated, but for know nothing is set in stone so I left this as is. Xen doesn't use this path.

include/grub/slr_table.h
krystian-hebel: But validity of the first entry is ensured by `grub_slr_init_table()` and code in `grub_slr_add_entry()`, what would `grub_slr_first_entry()` give on top of that?

krystian-hebel (commented):
@DemiMarie could you mark discussions as resolved wherever you're happy with the change?

docs/grub.texi
DemiMarie: I see.

grub-core/lib/i386/relocator32.S
DemiMarie: I see.  This really should have a comment somewhere, but adding it is the GRUB maintainers’ job, not yours.  Feel free to resolve this conversation.

include/grub/slr_table.h
DemiMarie: Nothing, consider this resolved.

include/grub/slr_table.h
DemiMarie: That falls into the “input is known trusted here” case.

include/grub/i386/msr.h
DemiMarie: For some reason I could not find this in my copy, probably because I was not looking at the right place.  Consider this resolved.

include/grub/i386/linux.h
DemiMarie: That’s fine, consider this resolved.

grub-core/mmap/mmap.c
DemiMarie: That’s fine, consider this resolved.

grub-core/loader/i386/linux.c
DemiMarie: It’s not about security, but about ease of readability and robustness to accedential changes.  That said, it’s not that big a deal.

grub-core/loader/i386/linux.c
DemiMarie: Not worth blocking merge over.

grub-core/loader/i386/slaunch.c
DemiMarie: This can be resolved.

grub-core/loader/i386/txt/acmod.c
DemiMarie: That’s fine, you can resolve this.

grub-core/loader/i386/txt/acmod.c
DemiMarie: I see.  Not worth blocking merge.

grub-core/loader/i386/txt/acmod.c
DemiMarie: That is weird.  This can be resolved.

include/grub/i386/slaunch.h
DemiMarie: I don’t remember why.  I suspect it was because I assumed this was passed to the MLE, but I don’t think it is.

DemiMarie (approved):
Looks good!

grub-core/loader/i386/txt/verify.c
```diff
@@ -0,0 +1,281 @@
...
+  txt_heap = grub_txt_get_heap ();
+  heap_base = grub_txt_reg_pub_read32 (GRUB_TXT_HEAP_BASE);
+  heap_size = grub_txt_reg_pub_read32 (GRUB_TXT_HEAP_SIZE);
+
+  if ( txt_heap == NULL || heap_base == 0 || heap_size == 0 )
+     return grub_error (GRUB_ERR_BAD_DEVICE,
+                 N_("TXT heap is not configured correctly"));
+
+  bios_size = grub_txt_bios_data_size (txt_heap);
+  /* We support versions >= 4, but bios_data->mle_flags is in versions >= 5. */
```
krystian-hebel: Great, `SNB_IVB_SINIT_20190708_PW.bin` is version 4 but it has `mle_flags`. ~Fu~ Thanks, Intel.

DemiMarie:
> @DemiMarie could you mark discussions as resolved wherever you're happy with the change?

GitHub isn’t allowing it, sadly.

DemiMarie (approved):

DemiMarie (approved):

DemiMarie (commented):
Looks okay

krystian-hebel:
Those two sneaky errors were breaking boot on Optiplex... I'll redo the same tests on HP just in case before clearing debug output and finally preparing patches for qubes-grub2

krystian-hebel:
No regression on HP, debug output cleared, fixes squashed to appropriate commits. Running per-commit build test (for bisectability) and will send patches tomorrow.

marmarek:
@DemiMarie can you send me signed message with commit ID which you approve (can be pasted clearsigned here, or via email)?

DemiMarie (approved):

SergiiDmytruk:
[Cleanup](https://github.com/TrenchBoot/trenchboot-issues/issues/48): @krystian-hebel, so this got upstreamed and we won't need this PR and branches anymore?

krystian-hebel:
> @krystian-hebel, so this got upstreamed and we won't need this PR and branches anymore?

FSVO upstreamed, it was implemented in Qubes GRUB repo. Can be closed.

Issue #11, opened 2023-09-30
Make `slaunch_module` reentrant


krystian-hebel:
Currently, `slaunch_module` doesn't handle multiple invocations very well. On subsequent calls, it overwrites pointer to previous SINIT without freeing it, leaking memory in the process. On boot only last module is used, given that it matches the platform.

Ideally, this function should fall back to previously loaded module if new one doesn't match the platform or is in older version; otherwise it should properly free memory used by previous module.

This would allow scripts in `grub.d` to iterate over SINITs in `/boot` and create `slaunch_module` entry for each one, removing from user the burden of manually copying just the proper SINIT.

SergiiDmytruk:
Closed by 13bf69f550024b02fb94fa4073c84643f4d91574, but GitHub decided to not do it.

PR #17, opened 2024-02-20
Add support for AMD SKINIT to slaunch

SergiiDmytruk:
This addresses TrenchBoot/trenchboot-issues#21 and #11.

SKL haven't been updated, so this can't be tested yet.

### DIFF
--- a/grub-core/loader/i386/skinit.c
+++ b/grub-core/loader/i386/skinit.c
@@ -35,6 +35,64 @@
 /* Offset to area for passing data to SKL. */
 #define SLB_PARAM(slb)    ((const grub_uint16_t *) (slb))[3]
 
+int
+grub_skinit_is_slb (const void *slb_base, grub_uint32_t slb_size)
+{
+  const grub_uint8_t skl_uuid[16] = {
+    0x78, 0xf1, 0x26, 0x8e, 0x04, 0x92, 0x11, 0xe9,
+    0x83, 0x2a, 0xc8, 0x5b, 0x76, 0xc4, 0xcc, 0x02,
+  };
+  /* We need space after SLB to pass SLRT to it. */
+  const grub_ssize_t max_size = GRUB_SKINIT_SLB_SIZE - SLRT_SIZE;
+
+  const grub_uint8_t *uuid;
+
+  if (slb_size > max_size)
+    {
+      grub_dprintf ("slaunch", "SLB is too large: %d > %d\n",
+                    slb_size, max_size);
+      return 0;
+    }
+
+  if (SLB_MEASURED (slb_base) > slb_size)
+    {
+      grub_dprintf ("slaunch", "SLB measured size is too large: %d > %d\n",
+                    SLB_MEASURED (slb_base), slb_size);
+      return 0;
+    }
+
+  if (SLB_ENTRY (slb_base) >= SLB_MEASURED (slb_base))
+    {
+      grub_dprintf ("slaunch", "SLB entry is not measured: %d >= %d\n",
+                    SLB_ENTRY (slb_base), SLB_MEASURED (slb_base));
+      return 0;
+    }
+
+  if (SLB_INFO (slb_base) > SLB_MEASURED (slb_base) - sizeof(skl_uuid))
+    {
+      grub_dprintf ("slaunch", "SLB info is not measured: %d > %d\n",
+                    SLB_INFO (slb_base),
+                    SLB_MEASURED (slb_base) - sizeof(skl_uuid));
+      return 0;
+    }
+
+  if (SLB_PARAM (slb_base) > max_size)
+    {
+      grub_dprintf ("slaunch", "SLB bootloader data offset is too large: %d > %d\n",
+                    SLB_PARAM (slb_base), max_size);
+      return 0;
+    }
+
+  uuid = (const grub_uint8_t *) slb_base + SLB_INFO (slb_base);
+  if (grub_memcmp (uuid, skl_uuid, sizeof(skl_uuid)) != 0)
+    {
+      grub_dprintf ("slaunch", "SLB has unexpected UUID\n");
+      return 0;
+    }
+
+  return 1;
+}
+
 grub_err_t
 grub_skinit_boot_prepare (struct grub_relocator *rel,
                           struct grub_slaunch_params *slparams)
--- a/grub-core/loader/i386/skinit.c
+++ b/grub-core/loader/i386/skinit.c
@@ -1,0 +1,99 @@
+/*
+ *  GRUB  --  GRand Unified Bootloader
+ *  Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ *  GRUB is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  GRUB is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with GRUB.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <grub/cpu/relocator.h>
+#include <grub/dl.h>
+#include <grub/i386/skinit.h>
+#include <grub/i386/tpm.h>
+#include <grub/mm.h>
+#include <grub/slr_table.h>
+#include <grub/types.h>
+
+#define SLRT_SIZE  GRUB_PAGE_SIZE
+
+/* Offset to entry point. */
+#define SLB_ENTRY(slb)    ((const grub_uint16_t *) (slb))[0]
+/* Amount of data actually measured for DRTM. */
+#define SLB_MEASURED(slb) ((const grub_uint16_t *) (slb))[1]
+/* Offset to structure with extra info. */
+#define SLB_INFO(slb)     ((const grub_uint16_t *) (slb))[2]
+/* Offset to area for passing data to SKL. */
+#define SLB_PARAM(slb)    ((const grub_uint16_t *) (slb))[3]
+
+grub_err_t
+grub_skinit_boot_prepare (struct grub_relocator *rel,
+                          struct grub_slaunch_params *slparams)
+{
+  grub_uint32_t *apic = (grub_uint32_t *)0xfee00300ULL;
+  const void *slb = grub_slaunch_module ();
+  grub_relocator_chunk_t ch;
+  grub_err_t err;
+  void *dce_mem;
+
+  if (slb == NULL)
+    return grub_error (GRUB_ERR_BAD_ARGUMENT, "SLB module is missing");
+
+  err = grub_relocator_alloc_chunk_align (rel, &ch, 0x1000000,
+                                          0xffffffff - GRUB_SLAUNCH_TPM_EVT_LOG_SIZE,
+                                          GRUB_SLAUNCH_TPM_EVT_LOG_SIZE, GRUB_PAGE_SIZE,
+                                          GRUB_RELOCATOR_PREFERENCE_NONE, 1);
+
+  if (err != GRUB_ERR_NONE)
+    return grub_error (err, "cannot alloc memory for TPM event log");
+
+  slparams->tpm_evt_log_base = get_physical_target_address (ch);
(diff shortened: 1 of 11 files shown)


grub-core/loader/multiboot_mbi2.c
```diff
@@ -1227,3 +1228,33 @@ grub_multiboot2_prepare_slaunch_txt (grub_uint32_t mbi_target,
 
   return GRUB_ERR_NONE;
 }
+
+grub_err_t
+grub_multiboot2_prepare_slaunch_skinit (grub_uint32_t mbi_target,
```
krystian-hebel: I think this should be merged with `*_txt()`, with differences inside one common function. `boot_params_addr` should be set also for Intel, otherwise we would have to handle it on kernel's side differently between Intel and AMD.

This may come at cost of one additional `memcpy` for SLRT, but it would be much easier to maintain in the long run.

grub-core/loader/i386/slaunch.c
```diff
@@ -227,13 +244,13 @@ grub_cmd_slaunch_module (grub_command_t cmd __attribute__ ((unused)),
 
   if (slp == SLP_INTEL_TXT)
```
krystian-hebel, line 248: We should have something for AMD, check for [UUID from skl_info](https://github.com/TrenchBoot/secure-kernel-loader/blob/master/main.c#L38) comes to mind, offset to which is in the [3rd word of file](https://github.com/TrenchBoot/secure-kernel-loader/blob/master/head.S#L37). Testing against `GRUB_SKINIT_SLB_SIZE` here also wouldn't hurt.

grub-core/loader/i386/skinit.c
```diff
@@ -0,0 +1,101 @@
...
+  grub_memcpy (dce_mem, grub_slaunch_module(), slparams->dce_size);
+
+  /* SLRT needs to be at least 4-byte aligned per specification */
+  dce_size_aligned = ALIGN_UP (slparams->dce_size, 4);
+  if (dce_size_aligned + GRUB_PAGE_SIZE > GRUB_SKINIT_SLB_SIZE)
+    return grub_error (GRUB_ERR_BUG, "not enough space for SLRT table");
+
+  slparams->slr_table_base = slparams->dce_base + dce_size_aligned;
+  slparams->slr_table_size = GRUB_PAGE_SIZE;
+  slparams->slr_table_mem = (grub_uint8_t *) dce_mem + dce_size_aligned;
```
krystian-hebel: I'm currently reworking SLK header format, so fields with offsets will be moved around. Most likely this will be something like this:

```
GLOBAL(sl_header)
	.word	_entry           /* SL header SKL offset to code start */
	.word	_end_of_measured /* SL header SKL measured length */
	.word	skl_info         /* Offset to SKL info with UUID and version */
	.word	bootloader_data  /* Offset to SLRT passed by the bootloader */
ENDDATA(sl_header)
```

First two words are defined by AMD, so they can't be redefined (although what we use them for may). `skl_info` is used in sanity check after build, but other than that it can be reordered if needed. `bootloader_data` used to be in place of `_end_of_measured`, but because of that whole SLK was measured, including BSS and data that could as well be built by SKL at runtime. SINIT sends SLK to the TPM for measuring, which is slow, so having it measure tens of KB of zeroes is a waste of time.

Memory layout than I'm aiming for is roughly:

- `sl_header`
- `.text` with all the code, `_entry` is somewhere within (currently at the very beginning, but doesn't have to be)
- `.rodata` with constants
- `skl_info` - it used to be in separate section, but perhaps it can be included in `.rodata`
- `_end_of_measured` - everything above is measured, constant and read-only, everything below isn't
- `.bss`, cleared by SKL because it can't trust GRUB (so `grub_memset` isn't strictly necessary here, except for SLRT)
- `bootloader_data` - this is where SLRT will live
- free space (hopefully)
- stack - `%esp` after SKINIT points to SLB+64K. We used to point it to stack located in `.text` but no need to measure it. SLK will check if stack is big enough, accounting for SLRT size, because it can't trust GRUB.
- there is no `.data`, everything that is currently there will have to be moved either to `.rodata` or `.bss`

So to sum up:
- `grub_skinit_get_sl_size()` may use `_end_of_measured` - everything else is initialized by SKL anyway
- `dce_size` isn't used anywhere but here (I think), but perhaps it should be always set to 64K
- `dce_size_aligned` (probably renamed) should be based on `bootloader_data`, and on SKL side it will be accessed directly, without manual alignment - linker script can assure that it will be 4B-aligned
- check for `(dce_size_aligned + GRUB_PAGE_SIZE > GRUB_SKINIT_SLB_SIZE)` is sufficient - SKL may change in a way that will change its stack requirements, and it should be able to handle this without having to update GRUB code.

grub-core/loader/i386/slaunch.c
```diff
@@ -77,6 +77,8 @@ grub_slaunch_init_slrt_storage (int arch)
   /* Setup DCE and DLME information. */
   slr_dl_info_staging.hdr.tag = GRUB_SLR_ENTRY_DL_INFO;
   slr_dl_info_staging.hdr.size = sizeof(struct grub_slr_entry_dl_info);
+  slr_dl_info_staging.bl_context.bootloader = GRUB_SLR_BOOTLOADER_GRUB;
+  slr_dl_info_staging.bl_context.context = slparams.boot_params_addr;
   slr_dl_info_staging.dce_base = slparams.dce_base;
   slr_dl_info_staging.dce_size = slparams.dce_size;
   slr_dl_info_staging.dlme_entry = mle_header->entry_point;
```
krystian-hebel, line 87: Base and size of DLME (aka. MLE) must also be passed to DCE, in addition to the entry point. In TXT, this information is passed through OS2SINIT, but there is no similar structure for AMD.

This of course will have to be reflected in the specification.

grub-core/loader/multiboot_mbi2.c
SergiiDmytruk: Merged: https://github.com/TrenchBoot/grub/compare/ad13737578f935bcff302407860c2c6da7f73dab..a2af951d4f61ac7a5ea819eba49ce979c544a5ff#diff-285279252527f5bc911013189ac09a0880271f60dc8f28f28fe68c2f5d517c74L1191

grub-core/loader/i386/slaunch.c
SergiiDmytruk: Added a bunch of sanity checks: https://github.com/TrenchBoot/grub/compare/ad13737578f935bcff302407860c2c6da7f73dab..a2af951d4f61ac7a5ea819eba49ce979c544a5ff#diff-ee6e12de6eb52b720e2e6f3792ff7228232d905b280bef353649ae579b9a110eR39

grub-core/loader/i386/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/ad13737578f935bcff302407860c2c6da7f73dab..a2af951d4f61ac7a5ea819eba49ce979c544a5ff#diff-33a2d35c8932ea248b7465a187fc4603651840c19cb98630f3d5276d9163d57bR85

Updated https://github.com/TrenchBoot/documentation/issues/23

grub-core/loader/i386/skinit.c
SergiiDmytruk: Took this into account:
1. https://github.com/TrenchBoot/grub/compare/ad13737578f935bcff302407860c2c6da7f73dab..a2af951d4f61ac7a5ea819eba49ce979c544a5ff#diff-ee6e12de6eb52b720e2e6f3792ff7228232d905b280bef353649ae579b9a110eR29
2. https://github.com/TrenchBoot/grub/compare/ad13737578f935bcff302407860c2c6da7f73dab..a2af951d4f61ac7a5ea819eba49ce979c544a5ff#diff-ee6e12de6eb52b720e2e6f3792ff7228232d905b280bef353649ae579b9a110eL76

> * `dce_size` isn't used anywhere but here (I think), but perhaps it should be always set to 64K

It's used in SLRT and is an accurate size of ACM in case of TXT, so used `_end_of_measured` for now.

grub-core/loader/i386/skinit.c
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+  if (SLB_ENTRY (slb_base) >= SLB_MEASURED (slb_base))
+    {
+      grub_dprintf ("slaunch", "SLB entry is not measured: %d >= %d\n",
+                    SLB_ENTRY (slb_base), SLB_MEASURED (slb_base));
+      return 0;
+    }
+
+  if (SLB_INFO (slb_base) > slb_size - sizeof(skl_uuid))
```
krystian-hebel:
```suggestion
  if (SLB_INFO (slb_base) > SLB_MEASURED (slb_base) - sizeof(skl_uuid))
```

SKl info should be in measured space.

SergiiDmytruk: Updated: https://github.com/TrenchBoot/grub/compare/a2af951d4f61ac7a5ea819eba49ce979c544a5ff..4543040551b7163496e651e19572bcf662ba42eb

grub-core/loader/i386/slaunch.c
```diff
@@ -227,31 +247,40 @@ grub_cmd_slaunch_module (grub_command_t cmd __attribute__ ((unused)),
...
-      if (!grub_txt_acmod_match_platform (slaunch_module))
+      if (!grub_txt_acmod_match_platform (new_module))
 	{
 	  grub_error (GRUB_ERR_BAD_FILE_TYPE, N_("SINIT ACM does not match platform"));
 	  goto fail;
 	}
     }
+  else if (slp == SLP_AMD_SKINIT)
+    {
+      if (!grub_skinit_is_slb (slaunch_module, size))
```
krystian-hebel:
```suggestion
      if (!grub_skinit_is_slb (new_module, size))
```

grub-core/loader/multiboot_mbi2.c
```diff
@@ -1214,16 +1227,19 @@ grub_multiboot2_prepare_slaunch_txt (grub_uint32_t mbi_target,
...
   grub_slaunch_finish_slr_table ();
 
   grub_dprintf ("multiboot_loader", "slr_table_base = %lx, slr_table_size = %x\n",
                 (unsigned long) slparams->slr_table_base,
                 (unsigned) slparams->slr_table_size);
-  grub_memcpy ((void *)(grub_addr_t) slparams->slr_table_base,
-               slparams->slr_table_mem,
-               slparams->slr_table_size);
+
+  if (slp == SLP_INTEL_TXT)
```
krystian-hebel, line 1239: This should happen for AMD also.

krystian-hebel (commented):
Few more problems discovered after starting on hardware. Still doesn't work after fixing them, but the rest of the problems is in SKL, I think.

include/grub/slr_table.h
```diff
@@ -119,6 +119,8 @@ struct grub_slr_entry_dl_info
   grub_uint64_t dl_handler;
   grub_uint64_t dce_base;
   grub_uint32_t dce_size;
+  grub_uint64_t dlme_base;
+  grub_uint32_t dlme_size;
   grub_uint64_t dlme_entry;
```
krystian-hebel: (Forgot to add this comment to the review)

Currently, this field specifies the offset from DLME base, not the physical address of entry point. As such, it will never be bigger than `dlme_size`, so it may as well be changed to `grub_uint32_t`.

SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/4543040551b7163496e651e19572bcf662ba42eb..13bf69f550024b02fb94fa4073c84643f4d91574#diff-fa8593325ef57d162ad9d9472e2e6571ec4dba071acb7da6e07515bbe8658a63

grub-core/loader/i386/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/4543040551b7163496e651e19572bcf662ba42eb..13bf69f550024b02fb94fa4073c84643f4d91574#diff-33a2d35c8932ea248b7465a187fc4603651840c19cb98630f3d5276d9163d57b

grub-core/loader/multiboot_mbi2.c
SergiiDmytruk: Why? `grub_skinit_boot_prepare()` assigns `slparams->slr_table_base` to physical and `slparams->slr_table_mem` to corresponding virtual address, there is no intermediate buffer for SLRT in this case.

krystian-hebel: You are correct. Not sure why I had to do this in testing, possibly `grub-install` failed between one version and another and I didn't notice.

krystian-hebel (approved):

PR #21, opened 2024-05-23
grub-core/loader/multiboot_elfxx.c: Fix calculating MLE size

miczyg1:
To avoid mismatch between OS SINIT data MLE size and MLE size from MLE header, take the mle_size from MLE header and do not align it to PMR size, which is 2MB.


miczyg1:
@krystian-hebel ping

krystian-hebel (approved):

PR #23, opened 2024-08-17
grub-core/loader/i386/txt/txt.c: Use MAXPHYADDR in MTRR masks

miczyg1:
Based on Intel TXT MLE Developer Guide revision 017.4 Table 4 the SINIT capabilities bit 8 indicates whether fixed 36bit masks or MAXPHYADDR masks are to be used in MTRR calculations. Failing to adhere to it may lead to creation of potentially disjoint WB cache ranges and violation of CRAM protections - according to the document.

### DIFF
--- a/grub-core/loader/i386/txt/txt.c
+++ b/grub-core/loader/i386/txt/txt.c
@@ -764,6 +764,15 @@ init_txt_heap (struct grub_slaunch_params *slparams, struct grub_txt_acm_header 
    */
   os_sinit_data->capabilities = GRUB_TXT_CAPS_TPM_12_AUTH_PCR_USAGE;
 
+  /* CBnT must set bits 4 and 5 */
+  if (sinit_caps & GRUB_TXT_CAPS_CBNT_SUPPORT)
+    {
+      os_sinit_data->capabilities |= GRUB_TXT_CAPS_CBNT_SUPPORT;
+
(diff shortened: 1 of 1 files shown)


krystian-hebel:
This shouldn't make a difference with current Xen implementation, in https://github.com/TrenchBoot/xen/blob/aem-4.17.4/xen/arch/x86/boot/head.S#L131 36-bit is selected, and according to Table 5. Truth Table of SINIT / MLE functionality of TXT SDG (Revision 017.4) it takes precedence. To make it work, we should also set this bit in Xen, but currently [there are too many branches](https://github.com/TrenchBoot/trenchboot-issues/issues/48) that would require this change.

miczyg1:
> This shouldn't make a difference with current Xen implementation, in https://github.com/TrenchBoot/xen/blob/aem-4.17.4/xen/arch/x86/boot/head.S#L131 36-bit is selected, and according to Table 5. Truth Table of SINIT / MLE functionality of TXT SDG (Revision 017.4) it takes precedence. To make it work, we should also set this bit in Xen, but currently [there are too many branches](https://github.com/TrenchBoot/trenchboot-issues/issues/48) that would require this change.

Yes, but then you end up with mixed MTRRs masks on a core. Some of them have MAXPHYADDR masks and some of them have 36bit masks... That's what I observed on VP6670

PR #27, opened 2024-12-11
loader/multiboot_mbi2.c: disable UEFI BS when booting with slaunch

krystian-hebel:
MLE (kernel or Xen) can't use pre-DLE services, as those are not measured as part of DRTM. To make sure this is true, disable boot services before MBI is created, which will ensure that all required data is obtained and saved in MBI while it is available.

BS are disabled before call to grub_multiboot2_get_mbi_size(), so the amount of allocated memory will be sufficient.


SergiiDmytruk:
Closing in favor of merged https://github.com/TrenchBoot/grub/pull/28.  The approach here is simpler, but it doesn't work well in practice because neither Xen nor Linux in Dom0 expect to boot this way and end up confused trying to do a legacy boot on an EFI system.

PR #28, opened 2025-01-01
Enable booting of Slaunch-enabled EFI binaries via `chainloader` command

SergiiDmytruk:
Relevant changes are in the top 2 commits, the rest are more fixes noticed in the process (to be squashed later).


grub-core/loader/slaunch/x86_efi_linux.c
```diff
@@ -60,20 +60,44 @@ sl_efi_install_slr_table (struct grub_slaunch_params *slparams)
...
+      for (mle_hdr_offset = 0; mle_hdr_offset < 0x1000; mle_hdr_offset += 16)
+        {
+          mle_hdr = (struct grub_txt_mle_header *)((grub_addr_t)kernel_addr + mle_hdr_offset);
+          if (!grub_memcmp (mle_hdr->uuid, GRUB_TXT_MLE_UUID, 16))
+            break;
+        }
+
+      if (mle_hdr_offset >= 0x1000)
+        {
+          return grub_error (GRUB_ERR_BAD_ARGUMENT, N_("not slaunch kernel: no MLE header found"));
```
miczyg1:
```suggestion
          return grub_error (GRUB_ERR_BAD_ARGUMENT, N_("not a slaunch kernel: no MLE header found"));
```

grub-core/loader/slaunch/x86_efi_linux.c
```diff
@@ -205,36 +232,38 @@ grub_sl_efi_txt_setup (struct grub_slaunch_params *slparams, void *kernel_addr,
...
   slparams->platform_type = grub_slaunch_platform_type();
 
-  /* See comment in TXT setup function grub_efi_slaunch_setup_txt() */
-  slparams->boot_params = &boot_params;
-  slparams->boot_params_base = (grub_uint64_t)&boot_params;
+  /* See comment in TXT setup function grub_sl_efi_txt_setup () */
+  if (is_linux)
+    {
+      slparams->boot_params = &boot_params;
+      slparams->boot_params_base = (grub_addr_t)&boot_params;
```
miczyg1, line 340: I don't quite understand why we have two fields which seemingly hold the same data. Is there any difference between them?

miczyg1 (changes requested):

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: Three other errors lack article as well, so I did the same.  Can update all of them, but then shouldn't the article be "an"?  I think of it with "s" pronounced standalone (like "s launch") due to abbreviation.

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: `boot_params` holds virtual and `boot_params_base` physical address.  From `i386/linux.c`:
```c
    real_mode_mem = get_virtual_current_address (ch);

    if (grub_slaunch_platform_type () != SLP_NONE)
      {
        slparams.boot_params = real_mode_mem;
        slparams.boot_params_base = get_physical_target_address (ch);
      }
```

Why the suffix is `_base` and not `_pa` or something like that is not known to me :)  Secure Launch Spec also uses "base" at least once in the same sense.

grub-core/loader/slaunch/x86_efi_linux.c
miczyg1: I thought something seemed off with an 'a'. I forgot we pronounce "s-launch", so 'an' is probably the right choice.

grub-core/loader/slaunch/x86_efi_linux.c
miczyg1: I see. Well It would be good to have boot_params_va and boot_params_pa, because right know I cannot distinguish which one is which. I will start some discussion on Matrix. RIght now closing the thread, as it should not be blocking.

miczyg1: Ohh wait... These fields look like something GRUB-specific:

```
struct grub_slaunch_params
{
  grub_uint32_t boot_type;
  grub_uint32_t platform_type;
  struct linux_kernel_params *boot_params;
  grub_uint64_t boot_params_base;
```

I thought it is defined in the spec... I guess it would be better to have:

```
struct grub_slaunch_params
{
  grub_uint32_t boot_type;
  grub_uint32_t platform_type;
  grub_uint64_t boot_params_va;
  grub_uint64_t boot_params_pa;
```

For less confusion. As the slaunch is not Linux-specific (will need casts everywhere to linux_kernel_params *). @rossphilipson WDYT?

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/b2955865a398273bbe2d42b80f9af09f1756f389..35c725edd5a0aca6ba09ea4fe4976cb906f54399

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: Right, I mentioned the spec only because the reasoning for the suffix should be the same in both cases.

If changing a couple of the fields, then the rest should be changed too:
* `efi_memmap_mem` -> `efi_memmap_va`?
* `slr_table_base` -> `slr_table_base`
* `slr_table_mem` -> `slr_table_va`?
* `mle_start` -> `mle_pa`
* `mle_ptab_target` -> `mle_ptab_pa`
* `mle_ptab_mem` -> `mle_ptab_va`?
* `ap_wake_block` -> `ap_wake_block_pa`
* `dce_base` -> `dce_pa`
* `tpm_evt_log_base` -> `tpm_evt_log_pa`

Types are also inconsistent.  Addresses should probably all be `grub_addr_t`.

Which is why I didn't do it.  Lots of renames across lots of commits and review might result in yet other names.  `_target`, `_base`, `_mem` are also used as parameters and local variables.

krystian-hebel: > Addresses should probably all be `grub_addr_t`.

Not really. First of all, GRUB uses this for physical addresses only (not sure if that is what you meant by "all"), virtual addresses are either `void *` or pointer to proper type. This type is also defined differently between 64b and 32b GRUB. While it may be enough to be used internally by GRUB (it won't allocate above 4G in 32b anyway), fields in SLRT are defined to always be 64 bits.

grub-core/loader/efi/chainloader.c
```diff
@@ -45,13 +46,35 @@ GRUB_MOD_LICENSE ("GPLv3+");
...
+static grub_efi_uintn_t image_pages;
+
+static struct grub_slaunch_params slparams = {0};
+
+static void
+free_image (void)
+{
+  if (image_address)
+    {
+      grub_efi_system_table->boot_services->free_pages (image_address,
```
krystian-hebel, line 60: Isn't this done by `boot_services->unload_image`? It may result in something being freed twice with multiple `chainloader` entries in menu or when the `chainloader` is followed by a different boot protocol, which causes `grub_chainloader_unload()`.

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: I meant physical addresses.  Consistently using `grub_uint64_t` for them is a good option too.  Currently, `mle_start` and `ap_wake_block` are of type `grub_uint32_t`.

grub-core/loader/efi/chainloader.c
SergiiDmytruk: No, see https://uefi.org/specs/UEFI/2.10_A/07_Services_Boot_Services.html#efi-boot-services-loadimage:

> If SourceBuffer is not NULL, the function is a memory-to-memory load in which SourceBuffer points to the image to be loaded and SourceSize indicates the image’s size in bytes. In this case, the caller has copied the image into SourceBuffer and **can free the buffer once loading is complete**.

So it always makes a copy.

krystian-hebel: Ah, I was so focused on searching any info about it in UnloadImage description that I didn't bother to check LoadImage :facepalm:

grub-core/loader/slaunch/skl.c
```diff
@@ -98,7 +98,7 @@ grub_skl_set_module (const void *skl_base, grub_uint32_t size)
   if (module->skl_info_offset > module->length - sizeof (info->uuid))
     {
       grub_dprintf ("slaunch",
-                    "Possible SKL module doesn't measure info: %u > %u\n",
+                    "Possible SKL module doesn't measure info: %u > %lu\n",
```
miczyg1, line 101: This actually breaks the build for i386-pc:

```
In file included from ../../include/grub/disk.h:31,
                 from ../../include/grub/file.h:26,
                 from ../../include/grub/loader.h:23,
                 from ../../grub-core/loader/slaunch/skl.c:35:
../../grub-core/loader/slaunch/skl.c: In function ‘grub_skl_set_module’:
../../grub-core/loader/slaunch/skl.c:101:21: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 6 has type ‘unsigned int’ [-Werror=format=]
  101 |                     "Possible SKL module doesn't measure info: %u > %lu\n",
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  102 |                     module->skl_info_offset,
  103 |                     module->length - sizeof (info->uuid));
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                    |
      |                                    unsigned int
../../include/grub/misc.h:38:88: note: in definition of macro ‘grub_dprintf’
   38 | #define grub_dprintf(condition, ...) grub_real_dprintf(GRUB_FILE, __LINE__, condition, __VA_ARGS__)
      |                                                                                        ^~~~~~~~~~~
../../grub-core/loader/slaunch/skl.c:101:71: note: format string is defined here
  101 |                     "Possible SKL module doesn't measure info: %u > %lu\n",
      |                                                                     ~~^
      |                                                                       |
      |                                                                       long unsigned int
      |                                                                     %u
```

miczyg1:
@SergiiDmytruk can we get packages built by CI here too?

SergiiDmytruk:
> @SergiiDmytruk can we get packages built by CI here too?

I'm trying to make CI work at https://github.com/TrenchBoot/grub/commits/xen-efi-ci/.

grub-core/loader/slaunch/skl.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/35c725edd5a0aca6ba09ea4fe4976cb906f54399..6965858d99a49f207be546f69289de711ad33538 should fix it.

macpijan:
@miczyg1 can you please resolve your threads ?

grub-core/loader/slaunch/x86_efi_linux.c
miczyg1: If these are GRUB internal fields then GRUB may use its own types it sees fit (and can fit the values the intend to hold).

But if it is SLRT then it must be a fixed size, 64 bit for all addresses.

> If changing a couple of the fields, then the rest should be changed too:
> Types are also inconsistent

Your comments just confirm that it is a little bit messy. Names are one thing, types are another. There seem to be a pattern:

- `_base` - means PA
- `_mem` - means VA

Exception: `mle_ptab_target`  is PA and has its `_mem` counterpart (`mle_start` and `ap_wake_block` are PA only, so I guess it is fine). So maybe at a minimum let's use `boot_params_mem` for VA?

To be consistent we would need also `mle_ptab_target`-> `mle_ptab_base`? But I wonder how intuitive it is. Leaving it as open question.

miczyg1 (approved):
From my side only the field names are to be hanged. As it is just a minor thing, I already approve the rest.

grub-core/loader/slaunch/x86_efi_linux.c
```diff
@@ -99,6 +99,16 @@ sl_efi_locate_mle_offset (struct grub_slaunch_params *slparams,
...
   slparams->mle_entry = mle_hdr->entry_point;
 
+  if (!is_linux) {
+    /*
+     * The previous value of the field is covering the whole EFI image which
+     * can include a lot of useless padding.  Limit the size used for measuring
+     * MLE to that reported by the header.  Don't change the behaviour for
+     * Linux.
+     */
+    slparams->mle_size = mle_hdr->mle_end - mle_hdr->mle_start;
```
krystian-hebel, line 109: This is related to https://github.com/TrenchBoot/xen/pull/21#discussion_r1961703776. I think that field is used to check if PMR ranges are correct, so it should be made valid in the header.

krystian-hebel (commented):
Apart from one comment, the changes look good. I'm assuming that most of them will eventually be squashed into earlier commits, and because of that I didn't paid too much attention to commit descriptions and their relative order.

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: Interestingly, comparing `xen.efi.map` and `xen-syms.map`, it's not a 2 MiB difference but a bit over 3.5 MiB in my build.  However, it doesn't matter because linker works it out.  Sizes in headers of `xen` and `xen.efi` match symbol differences (`_end - start`) from corresponding map files.

Just to make it clear, entry points in headers don't differ just by 2 MiB, it's `0x428000` for my files.  `fixmlehdr` doesn't turn one header into another, linker resolves relocations as it should and yet `XEN_IMG_OFFSET` is not accounted for in `slaunch_stub_entry - start` in EFI case for some reason.  `xen.lds.S` and `efi.lds` have rather similar beginnings, so maybe linker just behaves differently in case of PE format.

SergiiDmytruk:
> I'm assuming that most of them will eventually be squashed into earlier commits, and because of that I didn't paid too much attention to commit descriptions and their relative order.

Right.  There are several branches on top of each other and I thought it will be easier to apply fixes and update all of them at some later point to not have unrelated branches now.  A couple commits at the top can be squashed/moved right now, I just left them above RC1 tag so far.

grub-core/loader/slaunch/x86_efi_linux.c
krystian-hebel: > Sizes in headers of xen and xen.efi match symbol differences (`_end - start`) from corresponding map files.

Yes, but shouldn't the size be `_end - (start - XEN_IMG_OFFSET)` under UEFI, to account for different base? Maybe I'm missing something, but I thought it looks like this currently:

```
                               size
                 /^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\
+----------------+----------------------------------+
|image base      |start                             |_end
+----------------+----------------------------------+
\__________________________________/
         measured part
```

where measured part size is short by that 2 MiB padding in the beginning.

krystian-hebel: Actually, the field in MLE header is `Offset within MLE binary file`, so neither of those is valid, and that field shouldn't be used this way. As the file is loaded (not verbatim) by UEFI that doesn't understand MLE format, there is no use for this field at all, so it may stay as it is.

krystian-hebel (approved):
Last commit adds few binary files (archived source code of dependencies), but AFAIU those are just for CI/qubesbuilder and won't be a part of final, hopefully upstreamed, code.

SergiiDmytruk:
> Last commit adds few binary files (archived source code of dependencies), but AFAIU those are just for CI/qubesbuilder and won't be a part of final, hopefully upstreamed, code.

Yes, maybe there is a prettier way of doing it (e.g., downloading files during CI build), but that would need investigation and after finally making CI work for the latest GRUB I didn't return to this.

PR #29, opened 2025-03-13
V0.5 rc2 updates

SergiiDmytruk:
This is meant to fix https://github.com/TrenchBoot/trenchboot-issues/issues/55.

### DIFF
--- a/grub-core/loader/efi/chainloader.c
+++ b/grub-core/loader/efi/chainloader.c
@@ -37,6 +37,7 @@
 #include <grub/i18n.h>
 #include <grub/net.h>
 #include <grub/slaunch.h>
+#include <grub/time.h>
 #if defined (__i386__) || defined (__x86_64__)
 #include <grub/macho.h>
 #include <grub/i386/macho.h>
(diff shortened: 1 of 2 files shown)


grub-core/loader/efi/chainloader.c
```diff
@@ -250,6 +251,57 @@ make_file_path (grub_efi_device_path_t *dp, const char *filename)
   return file_path;
 }
 
+/* Clear unused memory.  This is meant to make measurements of DLME more
+ * deterministic in the face of firmware not bothering to zero area containing
+ * padding. */
```
krystian-hebel: This means one of two things:
- DRTM measures padding that isn't used by Xen
- Xen uses memory that it doesn't initialize

We should at least mention which one it is as a known limitation.

.github/workflows/build.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - 'intel-txt-aem*'
       - 'xen-uefi'
+      - 'v0.5-*'
```
krystian-hebel: Is this expected to stay, or was it used just to get the packages from CI for testing?

grub-core/loader/efi/chainloader.c
SergiiDmytruk: This is for padding.  Xen can use uninitialized memory, but I don't know that it does, so not mentioning it.

https://github.com/TrenchBoot/grub/compare/f9b8f78b2fd7d2e945b53ddb3998c66a90cbbaed..a8d47f8f8e0f43fb8ca6276489ec74e83b4a78b0#diff-a532ce1960f728d4d196008e9beb0a0610b9aa7088dadfb233dc6c9b1ff1d782R404-R411

Also cleaned up code a bit.

.github/workflows/build.yml
SergiiDmytruk: Only to trigger CI.  Dropped: https://github.com/TrenchBoot/grub/compare/f9b8f78b2fd7d2e945b53ddb3998c66a90cbbaed..a8d47f8f8e0f43fb8ca6276489ec74e83b4a78b0#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721

krystian-hebel (approved):
I still don't like that the padding is measured, but it may be the best option we have right now.

SergiiDmytruk:
> I still don't like that the padding is measured, but it may be the best option we have right now.

We can't really do much about it unless measuring is done with some scatter-gather list (which, I think, has no chance of working with TXT).

Mind that this is not the padding at the end of the image, that one is not measured since https://github.com/TrenchBoot/grub/commit/1a750736ccace2fd62bc9e157802cfc0e24e47fa.  It's the padding in between PE sections (basically where there is `67 df d1 9f 3e b3 64 9d 6f 3f 66 31 e4 4d 0b d3 6b 8d 8d 19` hash in the second spoiler of https://github.com/TrenchBoot/trenchboot-issues/issues/55#issuecomment-2721764450).

PR #30, opened 2025-03-23
Fix AEM for Intel TXT on UEFI platforms

SergiiDmytruk:
See commit messages for explanation of the changes.

This fixes https://github.com/TrenchBoot/trenchboot-issues/issues/56

### DIFF
--- a/grub-core/loader/slaunch/x86_efi_linux.c
+++ b/grub-core/loader/slaunch/x86_efi_linux.c
@@ -114,7 +114,10 @@ sl_efi_load_mle_data (struct grub_slaunch_params *slparams,
                    sizeof (struct linux_kernel_info));
 
       if (OFFSET_OF (mle_header_offset, &kernel_info) >= grub_le_to_cpu32 (kernel_info.size))
-        return grub_error (GRUB_ERR_BAD_OS, N_("not an slaunch kernel: lack of mle_header_offset"));
+        {
+          grub_dprintf ("slaunch", "not an slaunch kernel: lack of mle_header_offset\n");
+          return GRUB_ERR_BAD_OS;
+        }
 
       mle_hdr_offset = grub_le_to_cpu32 (kernel_info.mle_header_offset);
       mle_hdr = (struct grub_txt_mle_header *)((grub_addr_t)kernel_addr + slparams->mle_header_offset);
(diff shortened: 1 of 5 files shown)


grub-core/loader/slaunch/x86_efi_linux.c
```diff
@@ -94,6 +94,12 @@ sl_efi_load_mle_data (struct grub_slaunch_params *slparams,
         {
           return grub_error (GRUB_ERR_BAD_ARGUMENT, N_("not an slaunch kernel: no MLE header found"));
         }
+
+  if (grub_memcmp (mle_hdr->uuid, GRUB_TXT_MLE_UUID, 16) != 0)
+    {
```
krystian-hebel: Opening bracket is added in a different commit than the closing one, and indentation looks suspicious, although the latter may be just GH rendering.

The comparison it performed few lines above for non-linux. It also uses different form of checking the result (`!grub_memcmp(...)` instead of `grub_memcmp(...) != 0`), but I believe your form is the one expected in GRUB. In any case, these should be at least consistent with each other.

grub-core/loader/slaunch/x86_efi_linux.c
```diff
@@ -58,15 +58,51 @@ sl_efi_install_slr_table (struct grub_slaunch_params *slparams)
...
+  sections = (struct grub_pe32_section_table *)
+      ((char *) coff_header + sizeof (*coff_header)
+       + coff_header->optional_header_size);
+
+  for (i = 0, section = sections; i < coff_header->num_sections; i++, section++)
+    {
+      grub_uint32_t fstart = section->raw_data_offset;
+      grub_uint32_t fend = fstart + section->raw_data_size;
+
+      if ((section->characteristics & 0x60) != 0 &&
```
krystian-hebel: Please use `GRUB_PE32_SCN_CNT_CODE` and `GRUB_PE32_SCN_CNT_INITIALIZED_DATA` for better readability, and maybe a variable for it to keep the line length reasonable.

krystian-hebel (changes requested):

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: 0727266682c09416ecc96260ac3f445ac6785eb2

grub-core/loader/slaunch/x86_efi_linux.c
SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/c14b586979254639803addccb1900f4e809353eb..b5517787efc2359908c268e7bad5eb0bd1043525#diff-f1a0470cbd9e974350e6314846b3b573917f4003a22ecbd928ef534518506ca8L119-L155

Commits should be fine now, don't see any issues with the formatting.

krystian-hebel (approved):

PR #33, opened 2026-04-29
.github: workflows: build-rc.yml: add

DaniilKl:
Related to this PR: https://github.com/TrenchBoot/.github/pull/16 .

### DIFF
--- a/.github/workflows/build-rc.yml
+++ b/.github/workflows/build-rc.yml
@@ -1,0 +1,30 @@
+name: Build and package QubesOS RC RPMs
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


DaniilKl:
I am closing this, as there were a decision to not use the access via tokens stored as GitHub secrets to the ci.3mdeb.com for security reasons. This is to prevent the tokens leak from GitHub, that could compromise the packages that will be signed and/or the entire ci.3mdeb.com instance. Reason: the Woodpecker does not have scoped access control for its tokens, every token give the full access to a Woodpecker instance.

